PR #9, opened 2018-10-05
Add tests WIP

### LINKED ISSUE #5: Add automated tests run in CI
@allan, we have have some Python tests that we'll be adding next, but we'd also like to add a step to run the current Reaktoro tests. What is your current "workflow" to run the tests locally? Do you use `ctest`?

alexandrebbruno:
fix #5


tests/Reaktoro/ActivityModels/TestAqueousActivityModelDebyeHuckel.cpp
```diff
@@ -21,51 +21,50 @@
...
-const Database db("supcrt98");
-
 TEST_CASE("Electrolyte Solution: NaCl")
 {
-    ChemicalEditor editor(db);
-    editor.addAqueousPhase("H2O(l) H+ OH- Na+ Cl-")
-        .setChemicalModelDebyeHuckel();
-
-    ChemicalSystem system(editor);
+	const Database db("supcrt98");
```
nicoddemus: 2 spaces for indentation? Was this intentional?

tests/python/equilibrium/test_equilibriumPath.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import pytest
+import pandas as pd
+import sys
+
+#pytest
+from pytest_regressions.plugin import num_regression
+
+#PyReaktoro
+from equilibriumProblemsSetup import *
```
nicoddemus: Is this to import the fixtures? This is not advisable, better to move the fixtures to a `conftest.py` file instead and drop this import altogether.

alexandrebbruno:
Hey @allanleal,

While I was trying to add some tests to the CI, I found that some of than are giving a different output each time that it is run 🤔 
Here are the link to results from travis and appveyor results: 
* https://ci.appveyor.com/project/Reaktoro/reaktoro/build/1.0.80/job/pxvgtlqilvrphi3h
* https://travis-ci.org/reaktoro/Reaktoro/builds/437704131?utm_source=github_status&utm_medium=notification

To solve that problem, I've just added evals to some eigen expressions, but, since I'm quite new to the code, could you give a closer look to the results? Just to check if they are physically meaningful

Reaktoro/Common/ChemicalVector.hpp
```diff
@@ -529,21 +529,21 @@ auto operator*(const ChemicalScalarBase<VL,NL>& l, const ChemicalVectorBase<VR,T
...
 {
-    return {diag(l.val) * r.val,
-            diag(l.val) * r.ddT + diag(r.val) * l.ddT,
-            diag(l.val) * r.ddP + diag(r.val) * l.ddP,
-            diag(l.val) * r.ddn + diag(r.val) * l.ddn};
+	return { eval(diag(l.val) * r.val),
+		eval(diag(l.val) * r.ddT + diag(r.val) * l.ddT),
+		eval(diag(l.val) * r.ddP + diag(r.val) * l.ddP),
+		eval(diag(l.val) * r.ddn + diag(r.val) * l.ddn) };
 }
```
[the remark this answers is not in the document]
alexandrebbruno: Hey @allanleal,

Sure, considering that something like that could happen, we've already tried to run the test without that change and they didn't pass either.

https://ci.appveyor.com/project/Reaktoro/reaktoro/build/1.0.80/job/pxvgtlqilvrphi3h
https://travis-ci.org/reaktoro/Reaktoro/builds/437704131?utm_source=github_status&utm_medium=notification

The biggest problem is that they aren't passing in the CI, even after that changes. Well, I will continue investigating that problem and looking for an answer :)

tests/python/equilibrium/equilibriumProblemsSetup.py
```diff
@@ -0,0 +1,216 @@
...
+    system = ChemicalSystem(editor)
+  
+    problem = EquilibriumInverseProblem(system)
+    problem.add(b"H2O", 1, b"kg")
+    problem.add(b"NaCl", 0.1, b"mol")
+    problem.fixPhaseVolume(b"Gaseous", 0.2, b"m3", b"CO2")
+    problem.fixPhaseVolume(b"Aqueous", 0.3, b"m3", b"1 kg H2O; 0.1 mol NaCl")
+    problem.fixPhaseVolume(b"Calcite", 0.5, b"m3", b"CaCO3")
+    
+    return (system, problem)
```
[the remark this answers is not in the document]
alexandrebbruno: Oh sure, 

A minor change in the binds fixed the necessite to use this "b" :)
The problem is that this change is only locally in my machine, as soon as I push than I'll remove that b. 

[ ] - push bind change and remove b

alexandrebbruno:
> @alexandrebbruno Great and comprehensive job! I have a few suggestions for **naming convention** and **file structure organization**:
> 
> 1. Regression tests are always made using python, so we could just have then under: `Reaktoro/tests/regression/` instead of `Reaktoro/tests/python`. This is to distinguish other types of tests we might introduce later (also using python).
> 2. Say we want to test `EquilibriumPath` class. We should name the file `test_EquilibriumPath.py` instead of `test_equilibriumPath.py`. Maybe **even better** (_for more uniform and consistent representation_), we name as `test_equilibrium_path.py` (only small case letters with underscores separating words). We could stick with this convention even when there are species names:
>    
>    * `test_kinetic_path_with_h2o_hcl_caco3_calcite_0_to_10_minutes.csv` instead of
>    * `test_kinetic_path_solver_kinetic_problem_with_H2O_HCl_CaCO3_Calcite_0_to_10_minutes_.csv`.
>    
>    > Note also some simplification in the file name.

Sure,

Those changes are fine by me :)

- [ ] change work directory structure and tests names

Reaktoro/Common/ChemicalVector.hpp
```diff
@@ -527,25 +541,46 @@ auto operator*(const ChemicalScalarBase<VL,NL>& l, const ChemicalVectorBase<VR,T
 {
     return r * l;
 }
+/*
+template<typename VL, typename TL, typename PL, typename NL, typename VR, typename TR, typename PR, typename NR>
```
nicoddemus: There is a lot of commented code, do we plan to remove it or to keep it around?

appveyor.yml
```diff
@@ -45,3 +44,5 @@ build_script:
   # DLL must be in PATH in order for `reaktoro` to be successfully imported
   - set PATH=%APPVEYOR_BUILD_FOLDER%\artifacts\bin;%PATH%
   - python ci\check_compiled_files.py
+test_script:
+  - pytest -n auto -ra -vv tests/python
```
nicoddemus: Use `--color=yes` here so AppVeyor also uses colors. 👍

tests/python/equilibrium/equilibriumProblemsSetup.py
```diff
@@ -0,0 +1,216 @@
+import pytest
+
+from reaktoro import *
```
nicoddemus: This is considered bad style in general

tests/python/equilibrium/equilibriumProblemsSetup.py
```diff
@@ -0,0 +1,216 @@
...
+
+#This file has some problems setup that can be used inside pytest
+#all fixtures return a tuple with the following structure: (system, problem)
+
+@pytest.fixture(scope='session')
+def equilibriumProblemSetupH2O_CO2_NaCl_Halite_60C_300P():
```
nicoddemus: As we discussed in person, I suggest to move these fixtures to a `conftest.py` file

tests/python/pythonTools.py
```diff
@@ -0,0 +1,85 @@
+import numpy as np
+from reaktoro import *
```
nicoddemus: Ditto for importing `*`

Reaktoro/Common/ChemicalVector.hpp
tadeu: Let's remove commented code please ;)

tests/Reaktoro/ActivityModels/TestAqueousActivityModelDebyeHuckel.cpp
tadeu: I think that @alexandrebbruno is using tabs

@alexandrebbruno, try to configure your editor to always use spaces (and change all tabs into spaces in this PR), otherwise the indentation gets mixed up

tests/python/equilibrium/equilibriumProblemsSetup.py
```diff
@@ -0,0 +1,216 @@
...
+    '''
+    Build a problem with 1kg of H2O, 100g of CO2 and 0.1mol of NaCl 
+    at 60ºC and 300 bar
```
tadeu: Minor, but please, change the `º` sign for the proper degrees sign (and add a space before units): `60 °C`, `1 kg`, 100 g`, `0.1 mol`

tests/python/equilibrium/equilibriumProblemsSetup.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.fixture(scope='session')
+def equilibriumProblemSetupH2O_CO2_NaCl_Halite_dissolved_60C_300P():
+    '''
+    Build a problem with H2O, H+, Na+, Cl-, HCO3-, CO2(aq), CO3-- and 
+    Halite at 60ºC and 300 bar 
+    '''
+    database = Database(b"supcrt98.xml")
+    
+    editor = ChemicalEditor(database)
+    editor.addAqueousPhase([b"H2O(l)", b"H+", b"OH-", b"Na+", b"Cl-", b"HCO3-", b"CO2(aq)", b"CO3--", b"CO(aq)"]) \
```
tadeu: Can this same thing be done without chaining dots? I think it would be better:

```python
aqueous_phase = editor.addAqueousPhase([b"H2O(l)", b"H+", b"OH-", b"Na+", b"Cl-", b"HCO3-", b"CO2(aq)", b"CO3--", b"CO(aq)"])
aqueous_phase.setActivityModelDrummondCO2()
gaseous_phase = editor.addGaseousPhase([b"H2O(g)", b"CO2(g)"])
gaseous_phase.setChemicalModelSpycherPruessEnnis()
...
```

tests/python/equilibrium/test_equilibriumSolver.py
```diff
@@ -0,0 +1,277 @@
+import pytest
+import numpy as np
+import sys
+import os
+
+# pytest
+from pytest_regressions.plugin import num_regression
+from _pytest.fixtures import fixture
+
+# PyReaktoro
```
tadeu: I think that these comments separating imports are not necessary

tests/python/equilibrium/equilibriumProblemsSetup.py
alexandrebbruno: hmmm, 
I get it. what do you think about just import reaktoro and use as follow?

`import reaktoro 

 database = reaktoro.Database(b"supcrt98.xml")`

Is it better? :)

tests/python/equilibrium/equilibriumProblemsSetup.py
alexandrebbruno: ;)

tests/python/equilibrium/test_equilibriumSolver.py
alexandrebbruno: ok :)

tests/Reaktoro/ActivityModels/TestAqueousActivityModelDebyeHuckel.cpp
alexandrebbruno: thanks Tadeu,
my msvc was using tab instead of space :p
Fixed

tests/python/equilibrium/equilibriumProblemsSetup.py
alexandrebbruno: Done it like we discussed :)

allanleal:
Alternatively:

~~~
import reaktoro as rkt

database = rkt.Database(b"supcrt98.xml")
~~~

Maybe we should use `rkt` as a recommended short notation for users 
(like `np` for `numpy`).

However, I think that "importing all" when scripting a small 
application/test file should not be so critical - and it is a lot more 
convenient. ;)

tests/python/pythonTools.py
```diff
@@ -0,0 +1,85 @@
...
+    
+    total_element_amounts = state.elementAmounts()
+    
+    pH = ChemicalProperty.pH(system)(properties).val
+    
+    elementAmountInPhase = []
+    for i in range(0,system.numPhases()):
+        elementAmountInPhase.append(state.elementAmountsInPhase(i))
+    
+    outputDic ={}
```
tadeu: It's better to call it just `output`

tests/python/pythonTools.py
```diff
@@ -0,0 +1,85 @@
...
+    outputDic['Density [kg/m³]'] = phase_densities
+    outputDic['Molar Volume [m³/mol]'] = phase_molar_volumes
+    outputDic['Volume Fraction [m³/m³]'] = phase_volume_fractions
+    outputDic['pH [-]'] = np.asarray([pH])
+    
+    return outputDic
+
+def TableToDictionary(table):
+    '''
+    @param pandas.DataFrame
```
tadeu: Please use PEP-8 style, rename the function to `convert_dataframe_to_dict`.
The docstring can also be removed (it becomes obvious from the function name).

Reaktoro/Common/ChemicalVector.hpp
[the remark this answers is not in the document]
alexandrebbruno: Hey @allanleal,

Great news!! Tests are passing without the evals :)
Like you said, the Eigen update that you've made fix this tiny problem. Good job 👍

tests/regression/python_tools.py
```diff
@@ -0,0 +1,76 @@
...
+    phase_masses = properties.phaseMasses().val
+    
+    phase_molar_volumes = properties.phaseMolarVolumes().val
+    
+    phase_volumes = properties.phaseVolumes().val
+    
+    phase_volume_fractions = phase_volumes/np.sum(phase_volumes)
+    
+    phase_densities = phase_masses/phase_volumes
+
```
allanleal: By defining `phase_volume_fractions` and `phase_densities` as above, we'll end up testing the right hand side computation in this test file. We should instead get these values from `properties`. There is no phase volume fractions in `ChemicalProperties` class, but there is `phaseDensities`:
~~~
phase_densities = properties.phaseDensities().val
~~~

tests/regression/python_tools.py
```diff
@@ -0,0 +1,76 @@
...
+    
+    elementAmountInPhase = []
+    for i in range(0,system.numPhases()):
+        elementAmountInPhase.append(state.elementAmountsInPhase(i))
+    
+    output ={}
+    output['Temperature [K]'] = np.asarray([T]) 
+    output['Pressure [Pa]'] = np.asarray([P])
+    output['Total element amount [mol]'] = np.asarray(total_element_amounts)
+    for i in range(0,system.numPhases()):
```
allanleal: Do we really need `np.asarray` here around vectors coming from C++ (Eigen vectors)? I think this was needed when we were working with Boost.Python. As far as I know, `pybind11` creates a shallow wrapper to Eigen vectors and matrices (meaning that this is possibly not needed).

tests/regression/python_tools.py
```diff
@@ -0,0 +1,76 @@
...
+    
+    return output
+
+def convert_dataframe_to_dict(table):
+
+    output = {}
+    for elem in table.head():
+        output[elem] = table[elem].astype(float).values
+    
+    return output
```
allanleal: @nicoddemus @tadeu This `pytest-regressions` plugin is really nice!

tests/regression/test_equilibrium_path.py
```diff
@@ -0,0 +1,59 @@
...
+    ):
+    '''
+    An integration test that checks result's reproducibility of 
+    the calculation of an equilibrium path between two states   
+    '''
+    
+    database = Database("supcrt98.xml")
+    
+    editor = ChemicalEditor(database)
+    editor.addAqueousPhase("H O C Na Cl")
```
allanleal: @alexandrebbruno I deleted an empty line here. Let me know if you are comfortable with me editing directly your files here on GitHub when it is just cosmetic stuff (or typos).

tests/regression/test_equilibrium_solver.py
```diff
@@ -0,0 +1,276 @@
...
+        'Eq-H2O, CO2, NaCl and Halite at 60 °C and 300 bar',
+        'Eq-H2O, CO2, NaCl and Halite already dissolved at 60 °C and 300 bar',
+        'Eq-H2O, Fe(OH)2, Fe(OH)3, NH3 and Magnetite',
+        'Inv-H, O, Na, Cl, Ca, Mg, C with fixed amount and activity',
+        'Inv-H, O, Na, Cl, Ca, Mg, C with defined pH',
+        'Inv-H, O, Na, Cl, Ca, C, Calcite with defined pH and fixed amount',
+        'Inv-H2O, CO2, NaCl, CaCO3, Calcite with fixed species mass',
+        'Inv-H2O, NaCl, CaCO3, CO2 with fixed mass, amount and alkalinity',
+        'Inv-H2H, NaCl, CaCO3, CO2, Calcite with fixed phase volume'
+        ]
```
allanleal: What about having more descriptive ids here? Say, `Eq => Equilibrium` and `Inv => Inverse Equilibrium`? If longer names are not OK, at least `Inv => InvEq`.

allanleal: @alexandrebbruno Just noticed these `ids` are then used in the naming of the generated `csv` files. For example:

`...oximate_overload_1_Inv_H__O__Na__Cl__Ca__C__Calcite_with_defined_pH_and_fixed_amount_.csv`

The commas `,` in the `ids` are causing a double-underline `__` in the file names. Not sure if this is intended. What about removing those commas? Is there a recommendation for not having these `ids` with the same name of the test files, for example:
~~~
equilibrium_problem_with_h2o_co2_nacl_halite_60C_300bar
~~~
?

python/PyReaktoro/Thermodynamics/Core/PyChemicalEditor.cpp
```diff
@@ -41,14 +41,20 @@ void exportChemicalEditor(py::module& m)
     auto addPhase2 = static_cast<GaseousPhase&(ChemicalEditor::*)(const GaseousPhase&)>(&ChemicalEditor::addPhase);
     auto addPhase3 = static_cast<MineralPhase&(ChemicalEditor::*)(const MineralPhase&)>(&ChemicalEditor::addPhase);
 
-    auto addAqueousPhase1 = static_cast<AqueousPhase&(ChemicalEditor::*)(std::vector<std::string>)>(&ChemicalEditor::addAqueousPhase);
-    auto addAqueousPhase2 = static_cast<AqueousPhase&(ChemicalEditor::*)(std::string)>(&ChemicalEditor::addAqueousPhase);
+    //Need to be this order so that pybind do not see a python string as C++ std::string not a C++ std::vector<std::string> 
+    //with each leter as an elemento of that vetor 
+    auto addAqueousPhase1 = static_cast<AqueousPhase&(ChemicalEditor::*)(std::string)>(&ChemicalEditor::addAqueousPhase);
+    auto addAqueousPhase2 = static_cast<AqueousPhase&(ChemicalEditor::*)(std::vector<std::string>)>(&ChemicalEditor::addAqueousPhase);
```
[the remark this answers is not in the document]
nicoddemus: AFAIU it was not actually a bug, @tadeu can explain it better. 😁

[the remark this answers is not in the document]
alexandrebbruno: @tadeu who came out with this workaround :)

@allanleal. It's seems to me that this problem has just been solved by a PR merged 5 hours ago, but for me it seems nice keep that change to make reaktoro work fine with users who try to compile reaktoro with an old version of pybind11. What do you think?

tests/conftest.py
```diff
@@ -0,0 +1,214 @@
...
+    '''
+    Build a problem with H2O, H+, Na+, Cl-, HCO3-, CO2(aq), CO3-- and 
+    Halite at 60 °C and 300 bar 
+    '''
+    database = Database("supcrt98.xml")
+    
+    editor = ChemicalEditor(database)
+    aqueous = editor.addAqueousPhase(["H2O(l)", "H+", "OH-", "Na+", "Cl-", "HCO3-", "CO2(aq)", "CO3--", "CO(aq)"])
+    aqueous.setActivityModelDrummondCO2()
+    gaseus = editor.addGaseousPhase(["H2O(g)", "CO2(g)"])
```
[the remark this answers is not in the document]
alexandrebbruno: great catch, thanks :)

tests/regression/test_equilibrium_solver.py
alexandrebbruno: Like we've discussed, I had to reduce those ids because pytest was having some trouble to find the .csv. But I agree that an id with Eq and Inv are quite confuse and could induce to some erros so, I will try to think in something to solve this :) 
Removing those "," could help us to get .csv with shorter names, thanks 👍 

> Is there a recommendation for not having these ids with the same name of the test files, for example:

There are no restriction to the ids name, they are used just to add a more detail information of the test, giving a more meaningful name. But in that case it could be nicer use the same name as the fixture, I will think about that :)

tests/regression/python_tools.py
alexandrebbruno: I've tried to check all the values that were released by ChemicalState::output(std::string filename), but your argument make total sense :)
It makes much more sense use the phaseDensities for that case

python/PyReaktoro/Thermodynamics/Core/PyChemicalEditor.cpp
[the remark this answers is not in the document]
allanleal: 100% agree with you @alexandrebbruno. Please keep your changes. The PR on pybind11 you mention was actually requested by me and another guy who helped me with passing the CI tests.

tests/regression/test_equilibrium_solver.py
allanleal: +1

alexandrebbruno:
> Hi @alexandrebbruno , just check if we indeed need np.asarray wrapping Eigen vectors (which pybind11 actually converts to numpy views) and you are good to go in my opinion! Thanks a lot.

Hey, I remember that I tried to make those tests without the `np.asarray` but I end up with problems.
I really thought that I had mention this in your first comment but it seems that I end up going to lunch forgot to answer, really sorry. My bad :/

allanleal:
No worries! ;)

PR #11, opened 2018-10-08
Add Invoke tasks.py with first task to generate MSVC project (WIP)

tadeu:
Add more configurations for development mode environment (what is used to run tests in CI and local development)
Fix yaml-cpp lib configuration so that it works for Release and Debug without needing to regenerate MSVC project


tasks.py
```diff
@@ -0,0 +1,72 @@
+from invoke import task
```
nicoddemus: Can we get a new `Developing Quick-Start` section in the README explaining how to set this up for development? I mean a quick step-by-step one liners on how to start, something [like this step-by-step](https://github.com/ESSS/pytest-regressions/blob/master/HOWTORELEASE.rst) we have for releasing a new package.

tasks.py
```diff
@@ -0,0 +1,72 @@
...
+import shutil
+import sys
+
+
+def strip_and_join(s):
+    return ' '.join(line.strip() for line in s.splitlines() if line.strip() != '')
+
+
+def echo(c, msg):
+    from colorama.ansi import Fore, Style
```
nicoddemus: This probably should be at the top level

tasks.py
```diff
@@ -0,0 +1,72 @@
+from invoke import task
+from pathlib import Path
```
nicoddemus: This won't work on Python 2 unfortunately, we need `pathlib2` for that.

tasks.py
```diff
@@ -0,0 +1,72 @@
...
+        CMAKE_GENERATOR = "Visual Studio 14 2015"
+        CMAKE_ARCH = "x64"
+        CONFIG = "Release"
+        CMAKE_INCLUDE_PATH = f"{os.environ['CONDA_PREFIX']}\\Library\\include"
+
+        # `PYTHON_INSTALL_PREFIX` is configured to `artifacts_dir / 'python'`
+        # so that it won't "pollute" the Python environment when in develop
+        # mode.
+
+        c.run(strip_and_join(f"""
```
nicoddemus: f-strings won't work on Python 2 as well

tasks.py
```diff
@@ -0,0 +1,72 @@
...
+            activate reaktoro
+        """
+        root_dir = Path(__file__).parent
+        build_dir = root_dir / "build/msvc"
+        if clean:
+            remove_directory(c, build_dir)
+        build_dir.mkdir(parents=True, exist_ok=True)
+        os.chdir(build_dir)
+
+        relative_root_dir = Path(os.path.relpath(root_dir, Path.cwd()))
```
nicoddemus: There's `Path.relative_to` you can use here

tasks.py
tadeu: Hmm, should we support Python 2?

tasks.py
tadeu: Ok, for now it would be something like:

1) Install Miniconda
2) Execute `conda devenv`, `source activate reaktoro`
3) Execute `inv compile` to build it from sources (or `inv msvc` to creat an MSVC project on Windows)

do you agree?

tasks.py
tadeu: I like to keep imports local when they are only used in a single place, is there a problem to keep them there and move to global if later another place needs it?

tasks.py
tadeu: I've tried that first, unfortunately it's not the same thing, `relative_to` only works when a directory is completely inside the other (i.e., it won't generate `../../` paths).

allanleal:
Hi @tadeu. Thanks for this. I'm OK with this approach of providing two mechanisms for developing with Reaktoro. The classic 100% cmake way as it already exists (i.e., cmake install all dependencies on the fly) and the conda way as you are proposing. I see you had some issues with the deps `yaml-cpp`. This lib is I guess not actually used, but there might be one or two header files including this lib. It will at some point be removed in favor of a headers-only json library. So, once @nicoddemus approves this, go ahead with the merge.

tadeu:
> Hi @tadeu. Thanks for this. I'm OK with this approach of providing two mechanisms for developing with Reaktoro. The classic 100% cmake way as it already exists (i.e., cmake install all dependencies on the fly) and the conda way as you are proposing. I see you had some issues with the deps yaml-cpp. This lib is I guess not actually used, but there might be one or two header files including this lib. It will at some point be removed in favor of a headers-only json library. So, once @nicoddemus approves this, go ahead with the merge.

@allanleal, I didn't know `yaml-cpp` was not used, thanks for the explanation. My next step (after packaging Reaktoro in conda-forge) indeed would be to try to package the external dependencies individually (including `yaml-cpp`). Good to know that this one won't be necessary :), I can also try to remove it later.

PR #28, opened 2018-11-19
Removing cotire

allanleal:
Removing cotire as it never actually improved the compilation time for Reaktoro. This is most likely due to the heavy dependency on Eigen template library. Cotire also brings some inconvenience - adding a translation unit to the project causes a full recompilation.


tadeu:
Nice! The change seems good, although it seems to have exposed a problem on the bindings that `cotire` was "hiding", a lot of tests are failing with this:

```
>       pH = ChemicalProperty.pH(system)(properties).val
E       TypeError: Unable to convert function return value to a Python type! The signature was
E       	(arg0: Reaktoro::ChemicalSystem) -> std::function<Reaktoro::ChemicalScalarBase<double,Eigen::Matrix<double,1,-1,1,1,-1> > __cdecl(Reaktoro::ChemicalProperties const & __ptr64)>
E       
E       Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
E       <pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
E       conversions are optional and require extra headers to be included
E       when compiling your pybind11 module.
tests\regression\python_tools.py:40: TypeError
```

Let's hope it's just a missing `#include`

allanleal:
> Let's hope it's just a missing #include

Yes, most likely this. Thanks for reporting these test failures. I'll fix the missing includes.

tadeu:
I'll try to take a look at this now

allanleal:
Sorry for the delay :)
Just included pybind11/stl.h in `PyChemicalProperty.cpp`.

tadeu:
Ahh ok, I hope you don't mind, but I've just added a new commit also putting some other #includes :)

allanleal:
Great - thanks, your changes are more complete. 

In my machine (Linux) the build is passing OK (even without the recent change today). Is the build failing under Windows?

tadeu:
Ahh, the compilation was fine before, it was the tests that were failing, it seems `pybind11` silently fails to export the function type, and then it will fail only on runtime

allanleal:
OK. I get it. How do I run the tests in my machine? `pytest .` from the root dir?

tadeu:
Yes!

allanleal:
Thanks for the help solving this issue!

PR #30, opened 2018-11-21
Fix flakiness in "kinetic solver" regression tests

### LINKED ISSUE #23: Flaky behavior while using to solve kinetic problems
Hi guys,

I've just found some interesting behavior while I was using reaktoro to solve some kinetic problems that I think it worth to be mention here to open a discussion about it.

Some of the kinetic tests are given different results depending the number of processors used by the PC. 

Here are some of the results that I've obtained:
* result from test_path_solve_final_state_2 using pytest -n0 
https://sftp.esss.com.br/~abarbosa/finalStaten0.txt

* result from test_path_solve_final_state_2 using pytest -n8 
https://sftp.esss.com.br/~abarbosa/finalStaten8.txt

* numerical difference
https://sftp.esss.com.br/~abarbosa/diff.txt

Primary investigation pointed out that it could be something related to the VCode library, which is used during the integration process at line 214 in ODE.cpp, but I'm not quite sure about it 🤔 

What do you guys think about it? Are there anyone who have face anything like that? What you think about the idea of replace VCode by another one? 🤔 

more information about the test can be found at https://github.com/alexandrebbruno/Reaktoro/tree/study-kinetic-flaky.

alexandrebbruno:
Hey guy,
This PR will finally fix those kinetic tests that have a flaky behavior when they are run in parallel! :D 

Just for curiosity, the main problem it was that the solve function change the value of the state, which were invoked once for pytest session so, that change on the state variable and the randomly of tests execution when run in parallel made than flaky :P

Fix: 
* Remove the state from the fixture and move it to be declared inside each test
* Change the scope of all fixture from session to function, by doing that we ensure that all fixtures will be called once per test not per session

fix #23


allanleal:
Hi @alexandrebbruno , good to know you manage to solve this and in the end it was just a simple thing to do, and not a critical fault in the ODE solver. Your changes looks good to me.

tests/regression/test_kinetic_solver.py
```diff
@@ -45,16 +49,15 @@ def kinect_problem_with_h2o_hcl_caco3_mgco3_co2_calcite():
...
 
-    state = equilibrate(problem)
-
-    state.setSpeciesMass("Calcite", 100, "g")
-
-    return (state, reactions, partition)
+    return (problem, reactions, partition)
 
 
-@pytest.fixture(scope="session")
```
nicoddemus: Oh definitely, I missed this apparently. Session-scoped fixtures ideally should be used for immutable state, unless if you are absolutely sure that no tests change that state. 👍

tests/regression/test_kinetic_solver.py
```diff
@@ -283,28 +297,35 @@ def test_kinetic_path_solve_complete_path(
...
+        state.setSpeciesMass(mineral.mineral_name, mineral.amount, mineral.unit)
 
     path = KineticPath(reactions)
 
     path.setPartition(partition)
 
     path.solve(state, time_span.ti, time_span.tf, time_span.unit)
 
-    stateDic = convert_reaktoro_state_to_dict(state)
+    state_dic = convert_reaktoro_state_to_dict(state)
```
nicoddemus: Might as well use `dict`, which is the official abbreviation for `dictionary`. 

You might consider also writing your own regression fixture for additional brevity:

```python
@pytest.fixture
def state_regression(num_regression):
    class StateRegression:
        def check(self, state):
            num_regression.check(convert_reaktoro_state_to_dict(state))
    return StateRegression()
```

And use it as:

```python
    state_regression.check(state)
```

nicoddemus:
> Remove the state from the fixture and move it to be declared inside each test

I don't see this in the code, is the comment valid?

tadeu:
@alexandrebbruno I think you'll have to generate the expected values again with `--force-regen` (to their final correct values now)

tests/regression/test_kinetic_solver.py
```diff
@@ -45,16 +49,15 @@ def kinect_problem_with_h2o_hcl_caco3_mgco3_co2_calcite():
...
-    state.setSpeciesMass("Calcite", 100, "g")
-
-    return (state, reactions, partition)
+    return (problem, reactions, partition)
 
 
-@pytest.fixture(scope="session")
-def kinect_problem_with_h2o_nacl_caco3_mgco3_hcl_co2_calcite_magnesite_dolomite_halite():
-
+@pytest.fixture(scope="function")
```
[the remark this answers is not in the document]
alexandrebbruno: yeh,
I had some trouble with it while I was doing those changes, and decided to put it here just to be more explicit and not get confused :P

alexandrebbruno:
> @alexandrebbruno I think you'll have to generate the expected values again with `--force-regen` (to their final correct values now)

working on it :)

tests/regression/test_kinetic_solver.py
alexandrebbruno: OMG, nice!! thanks!!!
working on it ;)

alexandrebbruno:
> > Remove the state from the fixture and move it to be declared inside each test
> 
> I don't see this in the code, is the comment valid?

Hey bruno,
it is basically on commit a515e0aa15a0429eca2b1c65ac343ecf861c9efb. 
:)

tadeu:
Thanks for the fix @alexandrebbruno :)

PR #32, opened 2018-11-22
Improvement in CMAKE and conda integration in Linux

allanleal:
This PR improves cmake + conda integration in Linux, and it introduces `ccache` automatic detection so that developers can easily benefit from caching compiled objects (implying much faster subsequent builds whenever there are no changes in the translation units/header files). This fits well when changing branches, or building from a new directory.

To benefit from `ccache`, only make sure `ccache` is installed, and the rest is automatic.


allanleal:
I'm going to need some help here to see the tests passing, because I'm getting a seg fault in my machine when running `python3.7 -m pytest .` under the project root dir.

CMakeLists.txt
```diff
@@ -2,29 +2,38 @@
 cmake_minimum_required(VERSION 3.9)
 
 # Set the name of the project
-project(Reaktoro)
+project(Reaktoro VERSION 1.0.0 LANGUAGES CXX)
 
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+# Set the cmake module path of the project
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```
allanleal: Using `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR` because is some user decides to handle Reaktoro dependency using cmake command `add_subdirectory(Reaktoro)`, then `CMAKE_SOURCE_DIR` is actually the path to the root dir of the user's source dir, not Reaktoro's source code. By using `PROJECT_SOURCE_DIR`, we fix this, which is set based on the `project` command scope.

CMakeLists.txt
```diff
@@ -2,29 +2,38 @@
...
 
-# Set the version of the project
-set(REAKTORO_VERSION_MAJOR "1")
-set(REAKTORO_VERSION_MINOR "0")
-set(REAKTORO_VERSION_MICRO "0")
-set(REAKTORO_VERSION "${REAKTORO_VERSION_MAJOR}.${REAKTORO_VERSION_MINOR}.${REAKTORO_VERSION_MICRO}")
+# Set the output directories of the built libraries and binaries
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR})
```
allanleal: `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_BINDIR` come from `GNUInstallDirs` module.

Also, `REAKTORO_VERSION_*` variables are not needed, because cmake already provides `PROJECT_VERSION_*` variables

CMakeLists.txt
```diff
@@ -2,29 +2,38 @@
...
-option(BUILD_INTERPRETER "Build the interpreter executable reaktoro." ON)
-option(BUILD_PYTHON      "Build the python wrappers and python package reaktoro." OFF)
-option(BUILD_TESTS       "Build tests." OFF)
+option(REAKTORO_BUILD_ALL         "Build everything." OFF)
+option(REAKTORO_BUILD_DEMOS       "Build demos." OFF)
+option(REAKTORO_BUILD_DOCS        "Build documentation." OFF)
+option(REAKTORO_BUILD_INTERPRETER "Build the interpreter executable reaktoro." ON)
+option(REAKTORO_BUILD_PYTHON      "Build the python wrappers and python package reaktoro." ON)
+option(REAKTORO_BUILD_TESTS       "Build tests." OFF)
+
```
allanleal: Add prefix `REAKTORO_` to the cmake variables.
Note that `REAKTORO_BUILD_PYTHON` is now `ON` by default.

CMakeLists.txt
```diff
@@ -43,23 +52,6 @@ if(LINK_PHREEQC)
...
-endif()
-
-# Find Boost
-find_package(Boost REQUIRED)
-
-# Set the output directories of the built libraries and binaries
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-
```
allanleal: Instead of using `lib` and `bin`, we now use `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_BINDIR`. In some systems, it should not be `lib`, but `lib64`, which is one example that shows why it is preferred to delegate to GNUInstallDirs to determine the right path.

CMakeLists.txt
```diff
@@ -92,24 +86,27 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
...
 endif()
 
-# Set the project directory names
-set(REAKTORO_DATABASES_DIR  ${CMAKE_SOURCE_DIR}/databases)
-set(REAKTORO_DEMOS_DIR      ${CMAKE_SOURCE_DIR}/demos)
-set(REAKTORO_DOCS_DIR       ${CMAKE_SOURCE_DIR}/docs)
-set(REAKTORO_EXTERNAL_DIR   ${CMAKE_SOURCE_DIR}/external)
-set(REAKTORO_RESOURCES_DIR  ${CMAKE_SOURCE_DIR}/resources)
-set(REAKTORO_SOURCE_DIR     ${CMAKE_SOURCE_DIR}/Reaktoro)
-
```
allanleal: This is not needed --most of these variables were never actually used.

Reaktoro/CMakeLists.txt
```diff
@@ -31,40 +28,38 @@ add_library(ReaktoroObject OBJECT ${HEADER_FILES} ${SOURCE_FILES})
 # Specify the third-party dependencies of Reaktoro to ensure proper build sequence
 add_dependencies(ReaktoroObject ${THIRDPARTY_TARGETS})
 
+# Enable automatic creation of a module definition (.def) file for a SHARED library on Windows.
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
+
```
allanleal: I know, `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` only makes sense when building under Windows so that this kind tells us to wrap it in a `if(MSVC)` branch, but since this is ignored anyway in other OSes, I think we can simplify and just use `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)` directly.

Reaktoro/CMakeLists.txt
```diff
@@ -31,40 +28,38 @@ add_library(ReaktoroObject OBJECT ${HEADER_FILES} ${SOURCE_FILES})
...
+    target_link_libraries(${lib} ${THIRDPARTY_LIBS})
+    set_target_properties(${lib} PROPERTIES OUTPUT_NAME Reaktoro)
+endforeach()
+
+# Install all library targets (shared and/or static libs)
+install(TARGETS ${REAKTORO_LIBS}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT libraries)
+
```
allanleal: The above code resolves the duplication that existed before. We now have a list of library targets, for which we later perform install, and other modifications (link, change name).

cmake/CondaRecognizer.cmake
```diff
@@ -0,0 +1,18 @@
+# This cmake module determines if a conda environment is active.
+# If so, CMAKE_INSTALL_PREFIX is set to the environmental variable CONDA_PREFIX.
+# This automatic behavior can prevented by passing the option -DCMAKE_CONDA_IGNORE=TRUE
+
+
+# Keep the previous value of CMAKE_INSTALL_PREFIX before a possible change to CONDA_PREFIX
+set(CMAKE_INSTALL_PREFIX_BEFORE_CONDA_PREFIX ${CMAKE_INSTALL_PREFIX})
+
+# Check if a conda environment is active
+if(DEFINED ENV{CONDA_PREFIX} AND NOT CMAKE_CONDA_IGNORE)
```
tadeu: I've added a suggestion [here](https://github.com/allanleal/reaktoro/pull/1) to change this behavior

python/reaktoro/CMakeLists.txt
```diff
@@ -1,46 +1,31 @@
...
-add_custom_command(
-    DEPENDS ${FILES} DEPENDS PyReaktoro
-    OUTPUT timestamp
-    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:PyReaktoro>
+# Create a custom target to build reaktoro python package during build stage
+add_custom_target(reaktoro ALL
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:PyReaktoro> 
         ${CMAKE_CURRENT_BINARY_DIR}/$<TARGET_FILE_NAME:PyReaktoro>
-    COMMAND ${CMAKE_COMMAND} -E touch timestamp
+    COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} --quiet install --prefix=${CMAKE_BINARY_DIR}
```
tadeu: Why is `setup.py` called here in this custom target and also on `install` below on line 27?
Also, why here it installs to `CMAKE_BINARY_DIR`, and there to `CMAKE_INSTALL_PREFIX`?

python/reaktoro/CMakeLists.txt
```diff
@@ -1,46 +1,31 @@
...
+set(SETUP_PY ${CMAKE_CURRENT_BINARY_DIR}/setup.py)
+
 # Configure the setup.py script with the provided CMake variables
 configure_file(${SETUP_PY_IN} ${SETUP_PY})
 
-# Collect all files in the reaktoro directory
-file(GLOB_RECURSE FILES ${CMAKE_CURRENT_SOURCE_DIR} *)
-
-# Define the install target
-if(NOT DEFINED PYTHON_INSTALL_PREFIX)
```
tadeu: Hmm, could we leave the possibility to override where `setup.py` installs things? The idea of this flag `PYTHON_INSTALL_PREFIX` was to be able to override and install in a "local" prefix when in develop mode (without mixing with things in the environment).  But of course, if it isn't specified, the default behavior could be as you changed: to install in `CMAKE_INSTALL_PREFIX`

allanleal: Sure, let's provide this flexibility here - I agree with you on the need for this. Let's use `REAKTORO_PYTHON_INSTALL_PREFIX` then.

python/reaktoro/CMakeLists.txt
tadeu: Ahhh, got it, we usually take a different approach, we point `PYTHONPATH` directly to the source (i.e., `<directory-where-project-was-cloned>/python` in case of Reaktoro), and also to the place where the binary module (`.so`/`.pyd`) is installed (this is why we allow to install it in a different place through `CMAKE_INSTALL_PREFIX` customization).

But I think it's ok this way also, this extra installation at build time won't hurt :)

.travis.yml
```diff
@@ -1,10 +1,12 @@
-language: minimal
+language: cpp
```
allanleal: I needed to change here to `cpp` to get `ccache` working with travis, as documented [here](https://docs.travis-ci.com/user/caching/#ccache-cache).

python/CMakeLists.txt
```diff
@@ -1,19 +1,12 @@
...
-    
-# Check if python was found
-if(NOT PYTHONINTERP_FOUND)
-    set(BUILD_PYTHON OFF)
-    message(WARNING "Python executable was not found. "
-        "Cannot build python wrappers and packages for Reaktoro.")
+# Check if a desired version for python has been provided
+if(NOT PYBIND11_PYTHON_VERSION AND NOT PYTHON_EXECUTABLE)
+    # If a prython version not given, use the default one in the system 
+    find_package(PythonInterp REQUIRED) # PYTHON_EXECUTABLE variable is set here
```
allanleal: We need to use `find_package(Python)` as `PythonInterp` is now deprecated.

python/reaktoro/CMakeLists.txt
allanleal: In the past, I used to do the install to the binary dir, and the install target was of a `install(DIRECTORY ...)` type, instead of `install(CODE ...)` as it is now, to avoid double building. We can do this, or leave as is, because this specific build op is quite fast.

python/reaktoro/CMakeLists.txt
```diff
@@ -1,46 +1,38 @@
...
-       WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})"
+# Set the path where the python package is installed to CMAKE_INSTALL_PREFIX if not given
+if(NOT DEFINED REAKTORO_PYTHON_INSTALL_PREFIX)
+    set(REAKTORO_PYTHON_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
+endif()
+
+# Install the reaktoro python package using setuptools
+install(CODE
+"
+	file(TO_NATIVE_PATH \"${REAKTORO_PYTHON_INSTALL_PREFIX}\" REAKTORO_PYTHON_INSTALL_PREFIX_NATIVE)
```
allanleal: @tadeu Do we really need this command here (`file(TO_NATIVE_PATH `)?

tadeu: It's needed in Windows because for CMake we specify paths with forward slashes like `C:/Install` but when passing to `setup.py` it need to be in native format, like `C:\Install`

tadeu: > In the past, I used to do the install to the binary dir, and the install target was of a install(DIRECTORY ...) type, instead of install(CODE ...) as it is now, to avoid double building. We can do this, or leave as is, because this specific build op is quite fast.

Yes, it's very fast, I think it's ok to leave as it is 👍 

(weird that GitHub won't allow me to reply directly to this comment above)

python/CMakeLists.txt
tadeu: True, good to know that it's deprecated now. CMake version requirement will probably have to be updated to 3.12.
It seems there's a `find_package(Python3 COMPONENTS Interpreter)` now, specific for Python 3.

environment.devenv.yml
```diff
@@ -23,6 +24,9 @@ dependencies:
   - gxx_linux-64=7.3.0
   - libstdcxx-ng
 {% endif %}
+{% if sys.platform.startswith('linux') and sys.platform.startswith('osx') %}
```
tadeu: Should it be an `or` here?

allanleal:
Thanks @tadeu 

`ccache` is now working on travis (from 1000 seconds to 10!). 

I thought the `ccache` of the system would work, but it didn't (that's why the previous `CMAKE_CCACHE_IGNORE=TRUE`). I removed this, and now it is working like this:

- travis create the `.cache` dir (`/home/travis/.cache`), which survives for the next build
- `ccache` and `gcc` of miniconda is used together (instead of `ccache` and `gcc` of travis in `/usr/bin`)

It remains now to cache the conda part of the CI, which is taking relatively long (about 140 seconds only to start the env). Have you guys done this before? (cc @nicoddemus )

tadeu:
We "cache" conda envs in our internal Jenkins CI here, but I have never done this for Travis or Appveyor. If we are lucky someone will already have done that and we can try to follow the same steps, I did a quick search and found this discussion: https://github.com/theochem/qcgrids/issues/8

PR #35, opened 2018-12-07
fix travis and appveyor icons on readme file

alexandrebbruno:
hey guy,
I've just noticed that the path for travis and appveyor has not been changed to "/reaktoro" :)
I think that will fix the travis status icon, which were "unknown"


allanleal:
Thanks Alexandre!

PR #37, opened 2018-12-13
We can now use `find_package(Reaktoro)` to create C++ Reaktoro applications!

allanleal:
This PR introduces several changes that permit us now to use CMake `find_package(Reaktoro)` command to find Reaktoro's cmake targets (libraries, executables, etc). A demo called `reaktoro-app` is available in `Reaktoro/demos/cmake` directory (its README file explains how to build it). 

Here is the app (`main.cpp`):

~~~c++
#include <Reaktoro/Reaktoro.hpp>
using namespace Reaktoro;

int main()
{
    Database database("supcrt98.xml");

    ChemicalEditor editor(database);
    editor.addAqueousPhase("H O C Na Cl");
    editor.addGaseousPhase({"H2O(g)", "CO2(g)"});
    editor.addMineralPhase("Halite");

    ChemicalSystem system(editor);

    EquilibriumProblem problem(system);
    problem.setTemperature(60, "celsius");
    problem.setPressure(300, "bar");
    problem.add("H2O", 1, "kg");
    problem.add("CO2", 100, "g");
    problem.add("NaCl", 0.1, "mol");

    ChemicalState state = equilibrate(problem);

    state.output("state-cmake-example.txt");
}
~~~

and here is the `CMakeLists.txt` file:

~~~cmake
# Specify the minimum cmake version
cmake_minimum_required(VERSION 3.9)

# Specify the name of your project
project(reaktoro-app)

# Find the Reaktoro dependency.
find_package(Reaktoro REQUIRED)

# Create an executable named app with source code main.cpp
add_executable(app main.cpp)

# Link your application against Reaktoro C++ library.
target_link_libraries(app Reaktoro::Reaktoro)
~~~

It remains here to add a CI test to check if the file `state-cmake-example.txt` was successfully output by the app (so that we don't accidentally break this feature).


Reaktoro/CMakeLists.txt
```diff
@@ -9,57 +9,43 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
...
-
-if(LINK_PHREEQC)
-    set(THIRDPARTY_TARGETS ${THIRDPARTY_TARGETS} PHREEQC)
-    set(THIRDPARTY_LIBS ${THIRDPARTY_LIBS} phreeqc)
-endif()
-
-if(LINK_GEMS)
-    set(THIRDPARTY_TARGETS ${THIRDPARTY_TARGETS} GEMS)
-    set(THIRDPARTY_LIBS ${THIRDPARTY_LIBS} gems)
-endif()
```
allanleal: PHREEQC and GEMS backends always built from now on. There was never a reason not to do it, and this simplifies a bit the code.

Reaktoro/CMakeLists.txt
```diff
@@ -9,57 +9,43 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
...
+# Note: cmake v3.13 is needed to use target_link_directories command!
+link_directories(${REAKTORO_THIRDPARTY_LIBRARY_PATH})
 
-# Check if a static library for Reaktoro must be built
-if(BUILD_STATIC_LIBS)
-    add_library(ReaktoroStatic STATIC $<TARGET_OBJECTS:ReaktoroObject>)
-    list(APPEND REAKTORO_LIBS ReaktoroStatic)
-endif()
+# Build a Reaktoro C++ library
+add_library(Reaktoro ${HEADER_FILES} ${SOURCE_FILES})
```
allanleal: The user now needs to set `BUILD_SHARED_LIBS` to `OFF` if he wants a static library.

Reaktoro/CMakeLists.txt
```diff
@@ -9,57 +9,43 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
...
+add_dependencies(Reaktoro ${THIRDPARTY_TARGETS})
 
-# Install all library targets (shared and/or static libs)
-install(TARGETS ${REAKTORO_LIBS}
+# Add the include paths to the Reaktoro target
+target_include_directories(Reaktoro
+    PUBLIC
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+    PRIVATE ${REAKTORO_THIRDPARTY_INCLUDE_PATH})
```
allanleal: I know, this cmake part here is a bit convoluted, but it is the cmake-way to tell what are the include paths needed for building, and the include paths needed for users that are including Reaktoro.hpp in their apps.

Reaktoro/CMakeLists.txt
```diff
@@ -9,57 +9,43 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
...
+
+# Link Reaktoro library against external dependencies
+target_link_libraries(Reaktoro
+    PRIVATE ${THIRDPARTY_LIBS}
+    PUBLIC Boost::boost)
```
allanleal: Using a modern CMake way to link dependencies (`Boost::boost` target already contains all info on include paths).

Reaktoro/CMakeLists.txt
```diff
@@ -9,57 +9,43 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
...
+
+# Install Reaktoro C++ library
+install(TARGETS Reaktoro
+    EXPORT ReaktoroTargets
```
allanleal: This is needed for later installing a `ReaktoroTargets.cmake` file from which we can use the Reaktoro targets in downstream applications (e.g., `Reaktoro::Reaktoro` for the shared (or static) lib, `Reaktoro::ireaktoro` for the JSON interpreter).

Reaktoro/Math/LU.cpp
```diff
@@ -18,7 +18,7 @@
 #include "LU.hpp"
 
 // Eigen includes
-#include <Reaktoro/Math/Eigen/LU>
+#include <Reaktoro/deps/eigen3/Eigen/LU>
```
allanleal: Moved eigen from `Reaktoro/Math` to `Reaktoro/deps/eigen3`.

cmake/CondaRecognizer.cmake
```diff
@@ -12,6 +16,6 @@ if(DEFINED ENV{CONDA_PREFIX})
         set(CMAKE_INSTALL_PREFIX $ENV{CONDA_PREFIX})
     endif()
 
-    # Set BOOST_ROOT to CONDA_PREFIX to ensure conda's boost version is used
-    set(BOOST_ROOT $ENV{CONDA_PREFIX})
+    # Ensure dependencies from the conda environment are used instead of those from the system.
+    list(APPEND CMAKE_PREFIX_PATH $ENV{CONDA_PREFIX})
```
allanleal: Using `CMAKE_PREFIX_PATH` here is a much more general solution than `BOOST_ROOT`, which only affects Boost, and not the other dependencies that could exist on both conda environment and in the system.

thirdparty/CMakeLists.txt
```diff
@@ -45,22 +45,6 @@ ExternalProject_Add(PUGIXML
...
-    DOWNLOAD_NO_EXTRACT TRUE
-    DOWNLOAD_NO_PROGRESS TRUE
-    CONFIGURE_COMMAND ""
-    BUILD_COMMAND ""
-    INSTALL_COMMAND ${CMAKE_COMMAND} -E copy_directory
-        ${CMAKE_CURRENT_BINARY_DIR}/download/nlohmann
-            ${REAKTORO_THIRDPARTY_INSTALL_PREFIX}/include/nlohmann
-    CMAKE_ARGS ${REAKTORO_THIRDPARTY_COMMON_INSTALL_ARGS}
-)
-
```
allanleal: Json lib is no longer downloaded on the fly during build time, but allocated in the `Reaktoro/deps/json` dir. Reason: this simplifies a lot for Reaktoro users using the new `find_package(Reaktoro)` mechanism. They will not need to install json. This can be changed later when this json lib is available in conda-forge with equal versions for linux, windows and macos (linux version is much older).

Reaktoro/Math/LU.cpp
allanleal: This can be done. The issue is the following: currently, Reaktoro needs to use a non-release version of Eigen (dev version). Even the 3.3.7 version released recently does not have the features needed by Reaktoro (namely, indexed views). By explicitly using `<Reaktoro/deps/eigen3/Eigen/LU>` we ensure the embedded Eigen library is used, instead of a potentially conflicting Eigen library installed in the system.

Note this is a temporary solution. Once Eigen v3.4 is released (see last two comments [here](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=329)), we'll be able to access it from conda-forge and thus remove the internal eigen in Reaktoro.

PR #38, opened 2018-12-18
fix boundary conditions on transport solver

alexandrebbruno:
While I was using transportsolver class I released that it was not considering the boundary condition of the diffusion problem. This PR try to add this boundary condition to the problem. It also adds some documentation and tests
- [x] fix typos
- [x] add TODO on github by hand
- [x] remove second, minute, hour, year from demo-transport-solver.py and define day = 86400 # [ s ]
- [x] remove matplotlib from environment.devenv.yml
- [x] remove """ from test docs
- [x] remove 0 from range(0, nem_steps)
- [x] put ids next to values
- [x] change from reaktoro as rk to reaktoro as rkt


environment.devenv.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - boost=1.67
   - cmake>=3.9
   - invoke
+  - matplotlib
```
tadeu: Is the `demo` being automatically run in the CI jobs? If not, this wouldn't be necessary. Note that `matplotlib` is a heavy dependency, since it brings Qt.

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+    ],
+    ids=[
+        "diffusion-source rate q=0",
+        "diffusion-source rate q=1",
+        "diffusion-source rate q=x+1",
+        ],
+    )
+def test_transport_solver_diffusion(source_parameters, num_regression):
+    '''
+    """
```
tadeu: These `"""` should be removed

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+    A test to check the solution of a advection-diffusion equation with v = 0
```
nicoddemus: Excellent docs! 👍

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+    transp_solver.setDiffusionCoeff(D)
+    transp_solver.setBoundaryValue(ul)
+    transp_solver.setTimeStep(dt)
+      
+    transp_solver.initialize()
+      
+    numerical_u = np.zeros(num_cells)
+    transp_solver.step(numerical_u, q)
+      
+    for i in range(0,num_steps):
```
nicoddemus: `range(num_steps):` is enough, `range`'s first argument defaults to `0`

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+        transp_solver.step(numerical_u, q)
+    
+    num_regression.check({'u': numerical_u})
+
+@pytest.mark.parametrize(
+    "source_parameters",
+    [
+        (source_parameters(0, 0)),
```
nicoddemus: Fine to leave this as is, but is also possible to write this as:

```
    "source_parameters",
    [
        pytest.param(source_parameters(0, 0), id="advection-with source rate q=0"),
        pytest.param(source_parameters(0, 1), id="advection-with source rate q=1"),
        pytest.param(source_parameters(1, 1), id="advection-with source rate q=x+1"),
    ],
    ) 
```

This puts the ids next to the values, making them easier to keep in sync.

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    A test to check the solution of a advection-diffusion equation with v = 0
+    Eq: 
+        du/dt + v du/dx = D d²u/dx² + q
+    
+    The result were compared with the following analytic solution and got
+    near results when we increase the number of cells. We decided to compare with
+    numerical solve to save some computational time. 
+    
+    analytic_u = -((a*x**3)/(6*D)) - ((b*x**2)/(2*D)) + ((a*x*xr**2)/(2*D)) + ((b*x*xr)/(D)) + ul
```
tadeu: Do you have a reference for this? Usually the transient solution for pure diffusion is a gaussian (something like `u = a exp(-b (x - c)²) + d`).  Also, perhaps the analytical solution could be checked directly against the numerical solution, it should be super quick to evaluate.

tests/regression/test_transport_solver.py
tadeu: Indeed

Reaktoro/Transport.hpp
```diff
@@ -0,0 +1,20 @@
+// Reaktoro is a unified framework for modeling chemically reactive systems.
```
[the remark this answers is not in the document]
allanleal: This works as follows: if you need to include all `Core` stuff from Reaktoro, you could just include `Reaktoro/Core.hpp`. The same for `Transport.hpp`. However, there is only `TransportSolver.hpp` in the dir Transport.

Reaktoro/Transport/TransportSolver.cpp
```diff
@@ -266,9 +266,10 @@ auto TransportSolver::initialize() -> void
         A.row(icell) << a, b, c;
     }
 
+    // TODO: try to consider boundary with a quadradic error (unknown coefficients)
```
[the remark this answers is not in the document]
allanleal: Is there an automatic way to create issues from TODOs in the code? That would be nice.

Reaktoro/Transport.hpp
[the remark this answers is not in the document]
tadeu: Ahh got it, thanks for the explanation, it's for consistency ;)

Reaktoro/Transport/TransportSolver.cpp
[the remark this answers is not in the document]
tadeu: I don't know, but I just found this on Google: https://github.com/naholyr/github-todos :)

tests/regression/test_transport_solver.py
allanleal: One can also use the *method of manufactured solutions*, in which `u` is already known (e.g., `u(x ,t) = t * sin(x)`) from which the coefficients in the PDE is obtained (i.e., `v`, `D`, `q`).

environment.devenv.yml
allanleal: It's a good point - I don't think it is (and the plotting should not be needed in the CI).

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
+import numpy as np
+import pytest
+import reaktoro as rk
```
allanleal: Would you mind using `rkt` instead of `rk`? If you are already using `rk` in other tests, than that's OK.

Reaktoro/Transport/TransportSolver.cpp
```diff
@@ -266,9 +266,10 @@ auto TransportSolver::initialize() -> void
...
     // Assemble the coefficient matrix A for the boundary cells
-    A.row(icell0) << 0.0, 1.0 + beta, -beta;
-    A.row(icelln) << -beta, 1.0 + beta, 0.0;
+    A.row(icell0) << 0.0, 1.0 + 3*beta, -beta; // prescribed amount on the wall and approximatin deriveted by half control volume
+    A.row(icelln) << -beta, 1.0 + beta, 0.0;   // du/dx = 0 at the right boundary
```
allanleal: Do we always assume left boundary is `inlet` and right boundary `outlet`? What about using a backward 2nd order finite difference formula on the right boundary for the diffusion flux?

Reaktoro/Transport/TransportSolver.cpp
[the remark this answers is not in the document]
allanleal: Interesting - thanks!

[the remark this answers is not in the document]
alexandrebbruno: Interesting tool. For now I will just add those TODO on github by hand, but I think that we can work on that later

demos/python/demo-transport-solver.py
```diff
@@ -0,0 +1,58 @@
...
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+import reaktoro as rk
+
+# Auxiliary time related constants
+second = 1
+minute = 60
+hour = 60 * minute
+day = 24 * hour
```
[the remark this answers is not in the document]
alexandrebbruno: sure :)
It first I was using all this variables, but at the end I've just end up using only "day" :p

tests/regression/test_transport_solver.py
alexandrebbruno: It first I was checking directly against the analytical solution, but the problem is that was needing to increase the number o problem's cells up to 100 to get an error os 1e-2 :/ I this was taking to much time to solve the problem. So, I decided to compare against numerical solution

Sorry, I forgot to mention on the docs that I was solving the problem in steady state.

environment.devenv.yml
allanleal: Is it possible to use conda-devenv to specify that matplotlib should not be fetched in, say, Travis and AppVeyor? @alexandrebbruno  has an interesting point that this ensures that matplotlib is installed in the conda env on activated. But maybe we make this an exception and the user needs to install matplotlib explicitly.

tests/regression/test_transport_solver.py
```diff
@@ -0,0 +1,136 @@
...
+        ],
+    )
+def test_transport_solver_advection(source_parameters, num_regression):
+    '''
+    """
+    A test to check the solution of a advection-diffusion equation with D = 0
+    Eq: 
+        du/dt + v du/dx = D d²u/dx² + q
+    
+    The result were compared with the following analytic solution and got
```
jaimeambrus:
```suggestion
    The result was compared with the following analytic solution and got
```

tests/regression/test_transport_solver.py
alexandrebbruno: No problem :)
I can change it

tests/regression/test_transport_solver.py
tadeu: @alexandrebbruno, perhaps do it like this: keep the current number of cells and compare with the analytical solution using the relative tolerance that is needed (1e-1, it doesn't matter, but it's nice that it will show the error explicitly), but also continue using `num_regression` (because it is more sensitive to change, so it'll alert when the value changes if it shouldn't change).

Ah, also use an _absolute_ tolerance when comparing against the analytical solution, because the relative tolerance alone is bad when comparing values near zero (maybe this is why it's so high).

tadeu: @allanleal, nice that you mentioned the `method of manufactured solutions` 😃, we used this technique in the past (in other simulators), it's really very good, our reference is [this paper from Salari & Knupp (Sandia Labs)](https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2000/001444.pdf)

allanleal: thanks for the references @tadeu -- that method is quite helpful indeed. I would suggest @alexandrebbruno to just check if the calculations are roughly similar to the analytical solution and then use the computed values (not the analytical solution) for the regression. And @tadeu 's point on the need for absolute tol when comparing values close to zero is right on.

Reaktoro/Transport/TransportSolver.cpp
allanleal: OK - that suffices for now. At some point this transport module will get improved, but your additions now are super valuable.

environment.devenv.yml
allanleal: Clever stuff!

allanleal:
Great - thanks @alexandrebbruno !

PR #40, opened 2018-12-19
Pybind11 is now (temporarily) located in thirdparty dir

allanleal:
Some users have reported some runtime errors that I remember having in the past when using pybind11 without this [bugfix](https://github.com/pybind/pybind11/pull/1298).

pybind11 has not yet released a tag 2.3 with this bugfix, and conda-forge has v2.2.4 (without the bugfix).

This PR is a temp solution to ease the installation of reaktoro from sources without requiring the user to install pybind11 himself. Once v2.3 is available on conda-forge, this copy of pybind11 in the thirdparty dir can be removed.

------
**Updates**

Here are the requested changes:
- [x] change pythons' version back to 3.6
- [x] `ninja` is preferred to `make` when executing `cmake -P install`

Here are some additional changes you might want to know:

- cmake in `environment.devenv.yml`  has been updated to v3.13 (newest). Reason: not a strong or critical one, but the use of the command `cmake -S source-dir -B build-dir` to automatically create the build dir. With the use of `conda` now, one can easily abuse on the use of newest version for everything! :)


environment.devenv.yml
```diff
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = "3.6" %}
+{% set python_version = "3.7" %}
```
allanleal: Is there a reason why we use 3.6 instead of 3.7?

nicoddemus: We still use 3.6 here at ESSS because conda-forge has not yet generated packages for our entire stack in Python 3.7. I expect we migrate to Python 3.7 in the next couple of months. 👍

tadeu: Anaconda and conda-forge support the latest two minor versions of Python, so packages for Python 3.6 and 3.7 should be provided.  I left the CI at 3.6 since Python is usually forward-compatible at minor versions (leaving at 3.7 would have the risk of Reaktoro suddenly using 3.7-only features, breaking 3.6 packages).

tadeu: An alternative here would be to use the `CONDA_PY` environment variable so that the user can configure the Python version (with a default of `37`, but CI scripts would use `36`, or even a matrix building in both 3.6 and 3.7).

install
```diff
@@ -41,36 +41,36 @@ else()
...
 
-# Set the build type to Release by default for the external dependencies, if not provided
-if(NOT DEFINED THIRDPARTY_BUILD_TYPE)
-    set(THIRDPARTY_BUILD_TYPE Release)
-endif()
+# Extra arguments used during the build process (e.g., -j10)
+set(ARGS "")
+
+# If make installed, use make
+if(NOT DEFINED GENERATOR AND MAKE)
```
tadeu: `ninja` is usually faster than `make` (although not by much: https://david.rothlis.net/ninja-benchmark/). What do you think if `ninja` is the default in case both `make` and `ninja` are available?

environment.devenv.yml
allanleal: Thanks for the explanation - I'll roll back this to 3.6 and let @tadeu do this migration (or dual support) later as we discussed today.

install
allanleal: Sure. I'll make ninja de default if found (which should be the case in a conda env).

PR #41, opened 2018-12-21
Add Python 3.7 support in CI

tadeu:
and set it as default for `conda devenv`.

### DIFF
--- a/environment.devenv.yml
+++ b/environment.devenv.yml
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = "3.6" %}
+{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
 
 dependencies:
   - boost=1.67
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,12 @@ sudo: required
 
 cache: ccache
 
-matrix:
-  include:
-    - os: linux
+os:
+  - linux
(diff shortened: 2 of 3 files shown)


environment.devenv.yml
```diff
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = "3.6" %}
+{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
```
allanleal: Hi @tadeu . Any chance this `".".join(os.environ.get("CONDA_PY", "37"))` could get simpler?

allanleal: In Linux, env variables can contain dots:

```bash
$ export VAR=3.7
$ echo $VAR
3.7
````
Is this possible in Windows? If so, we can just set CONDA_PY to `3.7` or `3.6`.

tadeu: Ahh, in Windows it would also be possible, but I'm using `CONDA_PY` in this format in order to be standard with other tools that use this environment variable, such as `conda build`: https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html#environment-variables-that-affect-the-build-process

allanleal: OK... I suspected you had a strong reason for `36`, `37` and not `3.6` and `3.7`.

PR #42, opened 2018-12-21
Improve compilation in AppVeyor  using clcache

tadeu:


tadeu:
I just rebased this over #41, so that one should be merged first, then this one will only contain its own changes

environment.devenv.yml
```diff
@@ -29,6 +29,8 @@ dependencies:
 {% endif %}
 {% if sys.platform.startswith('linux') or sys.platform.startswith('osx') %}
   - ccache
+{% else %}
```
nicoddemus: You can remove this if/else now that `clcache` is used in all platforms. 😁

nicoddemus:
I triggered an AppVeyor rebuild to see how much `clcache` gains us.

tadeu:
😅 
Well, it's within the 1 Gb limit from the free tier. I'm thinking on caching the conda env next ;)

tadeu:
Down to 10 minutes each Windows job ;)

allanleal:
This is great @tadeu - thanks a lot!

PR #44, opened 2019-01-03
const on chemical system inside chemical state

armandoeng:


allanleal:
Hi @armandoeng . Is there a strong reason for this? If data member `system` of class `ChemicalSystem` becomes `const`, then the default constructor `ChemicalState()` needs to be deleted. I'm not sure yet the impact this could have. This change will enforce that a `ChemicalState` object is always initialized with a `ChemicalSystem` object, which could be desirable.

armandoeng:
> Hi @armandoeng . Is there a strong reason for this? If data member `system` of class `ChemicalSystem` becomes `const`, then the default constructor `ChemicalState()` needs to be deleted. I'm not sure yet the impact this could have. This change will enforce that a `ChemicalState` object is always initialized with a `ChemicalSystem` object, which could be desirable.

Hi @allanleal . In fact, it doesnt change any thing, because you dont have a `set function` for `system`.
So, right now, if you want to change a chemical system, you have to create a new chemical state, for what I can see.

tadeu:
@armandoeng, perhaps also try deleting the default constructor, as suggested by Allan:

```cpp
class ChemicalState
{
public:
    /// Disables the default constructor since it always needs a ChemicalSystem
    ChemicalState() = delete;
(...)
```

armandoeng:
Ok no problem

allanleal:
Looks good - just modified the doxygen comment of the deleted default constructor.

allanleal:
Sorry - closed by mistake. Please merge this @armandoeng !

Reaktoro/Interpreter/Interpreter.cpp
```diff
@@ -31,53 +31,53 @@ namespace Reaktoro {
...
+        for(auto item : node["substances"])
+            problem.add(item["substance"], item["quantity"], item["units"]);
+
+        ChemicalState state = equilibrate(problem);
+
+        std::string stateref = "default";
+        if(node.count("stateReference"))
+            stateref = node["stateReference"];
+
+        states.insert(std::map<std::string, ChemicalState>::value_type(stateref, equilibrate(problem)));
```
allanleal: Now I see. This and other cases made me decide to have `ChemicalState()` constructor. 
Compare:
```
states[stateref] = equilibrate(problem);
```
with
```
states.insert(std::map<std::string, ChemicalState>::value_type(stateref, equilibrate(problem)));
```
The former is a lot simpler and more convenient. I'm also concerned about this change in a reactive transport context, where we might want a `std::vector<ChemicalState>` and have a `lazy` initialization since we might not have the dimension yet.

armandoeng: I agree. I think its possible to have a const chemical system and a ChemicalState() constructor to supply these demands.

Reaktoro/Core/ChemicalQuantity.cpp
```diff
@@ -154,7 +154,7 @@ struct ChemicalQuantity::Impl
     ReactionSystem reactions;
 
     /// The chemical state of the system
-    ChemicalState state;
+    ChemicalState state {system};
```
allanleal: This has solved a compilation problem, but most likely will have an undesired runtime behavior because `state` will be initialized with an empty `system` instance.

Reaktoro/Interpreter/Interpreter.cpp
```diff
@@ -31,53 +31,53 @@ namespace Reaktoro {
...
-			problem.add(item["substance"], item["quantity"], item["units"]);
-
-		ChemicalState state = equilibrate(problem);
-
-		std::string stateref = "default";
-		if(node.count("stateReference"))
-			stateref = node["stateReference"];
-
-		states[stateref] = equilibrate(problem);
-	}
```
allanleal: Didn't notice I was using tabs instead of spaces here! :)

Reaktoro/Interpreter/Interpreter.cpp
allanleal: Yes, because of the `pimpl` idiom used in `ChemicalState`. So, yes, `const ChemicalSystem` inside `ChemicalState::Impl` while still keeping `ChemicalState()` constructor.

Reaktoro/Core/ChemicalQuantity.cpp
tadeu: In the previous code, was it also initializing it with an empty `system`, via the default constructor?

allanleal: Yes, that's right. But here, I think the right way to do this is via the constructor below (using member initialization list):

```
    /// Construct a custom ChemicalState::Impl instance
    Impl(const ChemicalSystem& system)
    : system(system), state(system)
    {
        // Initialise the molar amounts of the species and dual potentials
        n = zeros(system.numSpecies());
        y = zeros(system.numElements());
        z = zeros(system.numSpecies());
    }
```

Thus, we would not need the `ChemicalState state {system};`

tadeu: Ahh, true, it would be better ;)

armandoeng:
Some points to keep ChemicalState() = delete:

Instead of using: `states.insert(std::map<std::string, ChemicalState>::value_type(stateref, equilibrate(problem)));`
I could use: `states.insert({stateref, equilibrate(problem)});`

Instead of: `return pimpl->states.find(reference)->second;`
i could use: `return pimpl->states.at(reference);`
In this case, it would throw a exception if key doesnt exist. The operator[] would create the key.

allanleal:
@tadeu Not sure how the impact of this will be in python if we use a `dict` with `ChemicalState` values. Will we be able to do `states['InitialConditionState'] = state_ic`?

armandoeng:
> Do you have an application in which this change is essential for it to work? I would say we postpone this for now because, as I said in a recent comment, I'm not yet certain about the future impact this could have. For example, in reactive transport applications, we might want to have a std::vector<ChemicalState> data member which should be lazily initialized, as we might not have the dimension of that vector at construction time.

For vector, we could call reserve, which is more performatic, I think, and call emplace_back to make it to directly construct the object.

armandoeng:
> ot sure how the impact of this will be in python if we use a `dict` with `ChemicalState` values. Will we be able to do `states['InitialConditionState'] = state_ic`?

Well, I do not know that. I could make some tests. 
I think maybe it is risk to do that commit in the main reaktoro. We could do that commit only in our fork and test there.

Reaktoro/Core/ChemicalQuantity.cpp
```diff
@@ -177,19 +177,15 @@ struct ChemicalQuantity::Impl
...
 
-    /// Construct a default Impl instance
-    Impl()
-    {}
-
     /// Construct a custom Impl instance with given ChemicalSystem object
     Impl(const ChemicalSystem& system)
-    : system(system)
+    : system(system), state(system)
     {
```
allanleal: Hi @armandoeng I've made here the changes discussed with @tadeu .

Reaktoro/Core/ChemicalQuantity.cpp
```diff
@@ -177,19 +177,15 @@ struct ChemicalQuantity::Impl
...
     }
 
     /// Construct a custom Impl instance with given ReactionSystem object
     Impl(const ReactionSystem& reactions)
-    : system(reactions.system()), reactions(reactions)
+    : Impl(reactions.system()), reactions(reactions)
     {
```
allanleal: This other constructor also needs to initialize member `state` which is done using the other constructor call `Impl(const ChemicalSystem&)`.

Reaktoro/Core/ChemicalQuantity.cpp
armandoeng: Nice changes. Tnks.

allanleal:
> Some points to keep ChemicalState() = delete:
> 
> Instead of using: `states.insert(std::map<std::string, ChemicalState>::value_type(stateref, equilibrate(problem)));`
> I could use: `states.insert({stateref, equilibrate(problem)});`
> 
> Instead of: `return pimpl->states.find(reference)->second;`
> i could use: `return pimpl->states.at(reference);`
> In this case, it would throw a exception if key doesnt exist. The operator[] would create the key.

Sure, I like your suggestion. Please go ahead then with these alternatives.

Reaktoro/Core/ChemicalQuantity.cpp
armandoeng: I had to change it, because it was giving compilation problems. As you did call a constructor, it was constructing reactions inside it and it was not accpeting initializing here in this scope.

Reaktoro/Core/ChemicalQuantity.cpp
armandoeng: I just clarify the scope, to garantee that state was initialized with the system class member.

Reaktoro/Core/ChemicalQuantity.cpp
```diff
@@ -179,21 +179,22 @@ struct ChemicalQuantity::Impl
 
     /// Construct a default Impl instance
     Impl()
-    {}
+    : state(system)
```
armandoeng: To fix compilation error

Reaktoro/Core/ChemicalQuantity.cpp
allanleal: Great - this is actually better!

Reaktoro/Core/ChemicalQuantity.cpp
armandoeng: Done! My last commit for now. Have a good weekend. ;)

allanleal:
Thanks! Have a nice weekend!

allanleal:
Looks good @armandoeng ! :+1:

tadeu:
> @tadeu Not sure how the impact of this will be in python if we use a dict with ChemicalState values. Will we be able to do states['InitialConditionState'] = state_ic?

> Hi allan. I tested the dict on python with ChemicalState, works perfectly.

Good to know @armandoeng, could you add this as an unit test (if it's not there already)? Thanks! :)

tests/test_reaktoro_dicts.py
```diff
@@ -0,0 +1,43 @@
+import pytest
```
allanleal: I like @tadeu 's suggestion for a test case here. Shall we name this file as `test_reaktoro_with_python` or something else that is a bit more general than just `test_reaktoro_dicts`? This file will then serve as a place for testing Reaktoro using other Python constructs in the future.

armandoeng: Excelent idea.

PR #46, opened 2019-01-17
a more direct approach for adding species amounts

armandoeng:
Added function to set species using index, replicating the same idea of functions which sets elements by index.

### DIFF
--- a/Reaktoro/Equilibrium/EquilibriumProblem.cpp
+++ b/Reaktoro/Equilibrium/EquilibriumProblem.cpp
@@ -168,6 +168,30 @@ auto EquilibriumProblem::setElectricalCharge(double amount) -> EquilibriumProble
     return setElementAmount("Z", amount);
 }
 
+auto EquilibriumProblem::setSpecieAmount(const Species& species, double molar_amount) -> EquilibriumProblem&
+{
+    for(const auto& pair : species.elements())
+    {
+        const auto element = pair.first.name();
+        const auto coeffficient = pair.second;
+        const auto ielement = system().indexElement(element);
+        pimpl->b[ielement] += coeffficient * molar_amount;
+    }
+
+    return *this;
+}
+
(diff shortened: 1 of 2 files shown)


allanleal:
Hi @armandoeng . I think it makes sense only to be able to set temperature, pressure, and element amounts in `EquilibriumProblem`. The method `EquilibriumProblem::setSpeciesAmount` could be interpreted as the amount of given species being fixed at equilibrium. Even if we ignore this, you can achieve the same thing you want with the following code:

~~~python
state = ChemicalState(system)
state.setTemperature(300.0) # K
state.setPressure(10e5) # Pa
state.setSpeciesAmount('H2O(l)', 100.0);
state.setSpeciesAmount('CO2(aq)', 1.0);
state.setSpeciesAmount('Na+', 1.0);
state.setSpeciesAmount('Cl-', 1.0);

equilibrate(state)
~~~

You can also do:

~~~python
problem = EquilibriumProblem(system)
problem.add(state);
~~~

and now `problem` has collected the element amounts in `state`.

armandoeng:
Thnks.

armandoeng:
> 
> 
> Hi @armandoeng . I think it makes sense only to be able to set temperature, pressure, and element amounts in `EquilibriumProblem`. The method `EquilibriumProblem::setSpeciesAmount` could be interpreted as the amount of given species being fixed at equilibrium. Even if we ignore this, you can achieve the same thing you want with the following code:
> 
> ```python
> state = ChemicalState(system)
> state.setTemperature(300.0) # K
> state.setPressure(10e5) # Pa
> state.setSpeciesAmount('H2O(l)', 100.0);
> state.setSpeciesAmount('CO2(aq)', 1.0);
> state.setSpeciesAmount('Na+', 1.0);
> state.setSpeciesAmount('Cl-', 1.0);
> 
> equilibrate(state)
> ```
> 
> You can also do:
> 
> ```python
> problem = EquilibriumProblem(system)
> problem.add(state);
> ```
> 
> and now `problem` has collected the element amounts in `state`.

Thnks

PR #51, opened 2019-02-14
Use conda-devenv newly introduced preprocessing selectors to simplify dependency specs.

allanleal:
This PR mainly for discussion purposes on the simplification of `environment.devenv.py` in Reaktoro using [conda-devenv experimental support to preprocessing selectors](https://github.com/ESSS/conda-devenv/pull/79). This should also be a starting point for discussing other possible additions into conda-devenv to simplify even further.

If we all agree on the changes in conda-devenv and here, than we merge this PR at some point.

### DIFF
--- a/environment.devenv.yml
+++ b/environment.devenv.yml
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
+{% set py = env(PY_VER) or 3.7 %}
 
 dependencies:
   - boost=1.67
@@ -22,7 +22,7 @@ dependencies:
   - pytest-regressions
   - pytest-timeout
   - pytest-xdist
-  - python={{ python_version }}
+  - python={{ py }}
   - tabulate
   - gxx_linux-64=7.3.0             # [linux]
   - libstdcxx-ng                   # [linux]
@@ -32,12 +32,12 @@ dependencies:
 environment:
 
   PYTHONPATH:
-    - {{ root }}/artifacts/python/lib/python{{ python_version }}/site-packages   # [unix]
-    - {{ root }}/build/lib/python{{ python_version }}/site-packages              # [unix]
-    - {{ root }}/build/lib64/python{{ python_version }}/site-packages            # [unix]
+    - {{ root }}/artifacts/python/lib/python{{ py }}/site-packages   # [unix]
+    - {{ root }}/build/lib/python{{ py }}/site-packages              # [unix]
+    - {{ root }}/build/lib64/python{{ py }}/site-packages            # [unix]
     - {{ root }}\artifacts\python\Lib\site-packages                              # [win]
-    - {{ root }}\build\lib\python{{ python_version }}\site-packages              # [win]
-    - {{ root }}\build\lib64\python{{ python_version }}\site-packages            # [win]
+    - {{ root }}\build\lib\python{{ py }}\site-packages              # [win]
+    - {{ root }}\build\lib64\python{{ py }}\site-packages            # [win]
 
   LD_LIBRARY_PATH: {{ root }}/artifacts/lib # [unix]
 
--- a/environment.devenv.yml
+++ b/environment.devenv.yml
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set py = env('PY_VER') or 3.7 %}
+{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
 
 dependencies:
   - boost=1.67
--- a/environment.devenv.yml
+++ b/environment.devenv.yml
@@ -24,36 +24,21 @@ dependencies:
   - pytest-xdist
   - python={{ python_version }}
   - tabulate
-{% if sys.platform.startswith('linux') %}
-  - gxx_linux-64=7.3.0
-  - libstdcxx-ng
-{% endif %}
-{% if sys.platform.startswith('linux') or sys.platform.startswith('osx') %}
-  - ccache
-{% else %}
-  - clcache
-{% endif %}
+  - gxx_linux-64=7.3.0             # [linux]
+  - libstdcxx-ng                   # [linux]
+  - ccache                         # [unix]
+  - clcache                        # [win]
 
 environment:
-  {% if os.sys.platform != 'win32' %}
 
   PYTHONPATH:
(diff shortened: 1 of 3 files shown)


environment.devenv.yml
```diff
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
+{% set py = env(PY_VER) or 3.7 %}
```
allanleal: Here, we see the use case in which CONDA_PY (as an environment variable) is used to set the version of python of the conda environment.

The issue with `CONDA_PY` is its format adopted by conda-build: integer values such as 27 and 35 to denote semantic versions 2.7 and 3.5 respectively.

As we see below, it is cumbersome to convert `CONDA_PY` to a number with semantic version format:
~~~python
py = ".".join(os.environ.get("CONDA_PY", "37"))
~~~

With the new notation:

~~~python
py = env('PY_VER') or 3.7
~~~

we would need:

1. an `env` function in conda-devenv that resolves the value of the given environment variable, returning `None` if not set (or an empty string).
2. PY_VER, an environment variable, is set by conda-devenv on the fly, just like conda-build, based on the set value of `CONDA_PY`.

See [environment variables set during the build process with conda-build](https://docs.conda.io/projects/conda-build/en/latest/source/environment-variables.html?highlight=PY_VER#environment-variables-set-during-the-build-process). Check the `PY_VER` env variable.

Maybe there is an easier approach for simplifying this process of specifying python version in conda-devenv.

environment.devenv.yml
```diff
@@ -1,6 +1,6 @@
 name: reaktoro
 
-{% set python_version = ".".join(os.environ.get("CONDA_PY", "37")) %}
+{% set py = env('PY_VER') or 3.7 %}
```
nicoddemus: Probably using `"3.7"` (a string) instead of `3.7` (a `float`) is more correct, although I suspect the end result is the same. 👍

environment.devenv.yml
```diff
@@ -22,38 +22,23 @@ dependencies:
...
+    - {{ root }}/build/lib/python{{ py }}/site-packages              # [unix]
+    - {{ root }}/build/lib64/python{{ py }}/site-packages            # [unix]
+    - {{ root }}\artifacts\python\Lib\site-packages                  # [win]
+    - {{ root }}\build\lib\python{{ py }}\site-packages              # [win]
+    - {{ root }}\build\lib64\python{{ py }}\site-packages            # [win]
 
-  # Shared library must be in LD_LIBRARY_PATH in order for `reaktoro` to be successfully imported
-  LD_LIBRARY_PATH:
-    - {{ root }}/artifacts/lib
+  LD_LIBRARY_PATH: {{ root }}/artifacts/lib # [unix]
```
tadeu: Putting it in one line like this has a different meaning: it was a list before, now it's a single string. Consider using something like:

```
  LD_LIBRARY_PATH:  # [unix]
    - {{ root }}/artifacts/lib  # [unix]
```

environment.devenv.yml
tadeu: Agreed

environment.devenv.yml
allanleal: You're right that in general a string for version is preferred (e.g., 3.7.1 is not a number!). It should actually read as a string above ('3.7').

You're also right that the end result will be the same - jinja will convert (I assume) 3.7 to a string in each occurrence of `{{ py }}`.

environment.devenv.yml
allanleal: Yes, I was wondering about that - I was hoping that one could use either a list or a value for these environment variable sections. In case of a value, then interpreted as a list with a single value (or just set 

`export LD_LIBRARY_PATH={{ root }}/artifacts/lib:LD_LIBRARY_PATH`

allanleal: @tadeu Just tested here in a `environment.devenv.yml` file the following:

    environment:
        TESTENV: hellothere

After executing `conda devenv`, and then `conda activate myenv`, and finally:

    echo $TESTENV

it prints `hellothere`!

This is great, because we can then write one-liners for these environment variables.

@nicoddemus Can you comment if this is a loophole or intended behavior?

nicoddemus: Let me explain (we probably should make that more obvious in conda-devenv docs).

When you use the form:

```yaml
environment:
    TESTENV: hellothere
```

`conda-devenv` will **set** the environment variable to `hellothere`, overriding the previous contents (if any). This usage is meant for environment variables which have a single value, like `CONDA_PY` for example.

Whe you use the list form:

```yaml
environment:
    TESTENV: 
        - hellothere
```

Then `conda-devenv ` will **append** the `hellothere` value to any existing `TESTENV` environment variable, using the appropriate path separator (`:` on Linux, `;` on Windows). This is meant for environment variables which contain multiple values, like `PATH` and `PYTHONPATH`.

Hope that makes sense. 😁

allanleal: Crystal clear - so, @tadeu , we should go with your suggestion then:

~~~
  LD_LIBRARY_PATH:  # [unix]
    - {{ root }}/artifacts/lib  # [unix]
~~~

to not accidentally overwrite previous paths in `LD_LIBRARY_PATH`.

tadeu: 👍 
@nicoddemus, thanks for the explanation :)

nicoddemus: In that case it might be more readable to stick to use jinja blocks though, but of course is up to you. 😁

environment.devenv.yml
```diff
@@ -1,6 +1,6 @@
 name: reaktoro
```
nicoddemus: I suggest to add `min_conda_devenv_version("1.1")` here.

allanleal: Sure - will do.

allanleal:
@tadeu , what about we replace CONDA_PY as integer format to semantic
version format still in this PR? :)

tadeu:
> @tadeu , what about we replace CONDA_PY as integer format to semantic version format still in this PR? :)

Seems small enough to bundle it together :) 
I suggest the environment variable, tough, just as to not confuse with whoever uses `CONDA_PY`. Perhaps it could be `PY_VER`?

allanleal:
> Perhaps it could be `PY_VER`?

Good -  `PY_VER` then. Will do it later. Thanks!

allanleal:
Thanks @tadeu and @nicoddemus

PR #52, opened 2019-02-26
Enhanced units with additional entries commonly used in geochemistry

gdmiron:
Additional units and derived units used in (geo)chemistry. These entries were missing and are necessary for the cement application Dmitrii and I are working on.


Reaktoro/Common/Units.cpp
```diff
@@ -97,12 +97,20 @@ const double pi    = 3.14159265359;
...
+    {"permil"        , {{milli, "1", 1}}},
+    {"ppt"           , {{milli, "1", 1}}},
+    {"ppm"           , {{micro, "1", 1}}},
+    {"ppb"           , {{nano,  "1", 1}}},
 
     // Mass Units
     {"g"            , {{1, "g", 1}}},
     {"kg"           , {{kilo , "g", 1}}},
     {"mg"           , {{milli, "g", 1}}},
+    {"mkg"          , {{micro, "g", 1}}},
```
allanleal: `mk` for micro? The official symbol is µ, but since this is a Greek char, `u` is used instead (e.g., `ug` for micro gram). I propose we do not use this (and for the other occurrences below as well).

Reaktoro/Common/Units.cpp
```diff
@@ -97,12 +97,20 @@ const double pi    = 3.14159265359;
 map<string, DerivedUnit> derivedUnitsMap =
 {
     // Nondimensional Unit
-    {"1"            , {{1, "1", 1}}},
+    {"1"             , {{1, "1", 1}}},
+    {"%%"            , {{centi, "1", 1}}},
```
allanleal: Why not just one `%`?

Reaktoro/Common/Units.cpp
```diff
@@ -97,12 +97,20 @@ const double pi    = 3.14159265359;
...
+    {"percent"       , {{centi, "1", 1}}},
+    {"%%%%"          , {{milli, "1", 1}}},
```
allanleal: And here just two (i.e., `%%`)?

allanleal: The following code:

~~~cpp
#include <string>
#include <iostream>

int main(int argc, char const *argv[])
{
    std::string s1 = "%";
    std::string s2 = "%%";

    std::cout << s1 << std::endl;
    std::cout << s2 << std::endl;

    return 0;
}
~~~
outputs:
~~~
%
%%
~~~

Reaktoro/Common/Units.cpp
```diff
@@ -364,21 +407,28 @@ map<string, DerivedUnit> derivedUnitsMap =
...
+    {"neq"         , {{nano, "eq", 1}}}
 };
 
 map<string, TemperatureUnit> temperatureUnitsMap =
 {
     {"K"          , {1, 0, "K"}},
+    {"degK"       , {1, 0, "K"}},
     {"degC"       , {1, -273.15, "K"}},
     {"degF"       , {1.8, 32, "degC"}},
     {"degR"       , {1, 459.67, "degF"}},
```
allanleal: We should change `degR` to `R` for the same reasons above for K.

Reaktoro/Common/Units.cpp
```diff
@@ -364,21 +407,28 @@ map<string, DerivedUnit> derivedUnitsMap =
...
-    {"ueq"         , {{micro, "eq", 1}}}
+    {"ueq"         , {{micro, "eq", 1}}},
+    {"mkeq"         , {{micro, "eq", 1}}},
+    {"neq"         , {{nano, "eq", 1}}}
 };
 
 map<string, TemperatureUnit> temperatureUnitsMap =
 {
     {"K"          , {1, 0, "K"}},
+    {"degK"       , {1, 0, "K"}},
```
allanleal: °K is not correct, right? Because it is an absolute temperature scale, and the degree symbol is (in general) not used. We should not add `degK` then.

allanleal:
According to this article, we could use `mc` for micro (e.g., `mcg` for microgram).

Reaktoro/Common/Units.cpp
gdmiron: `u` for micro should be sufficient, if anything I see `mc` sometimes used for micro.

Reaktoro/Common/Units.cpp
gdmiron: the number of `%` is related to the power digit. `centi = 1.0e-2` therefore `%%`, but then `milli = 1.0e-3` should be `%%%`.

Reaktoro/Common/Units.cpp
gdmiron: If you want to stick only with the standard notation then its should be just `K`. This was more to have the non standard option.

Reaktoro/Common/Units.cpp
allanleal: I thought you were escaping `%` with `%` inside the string.

Anyway, isn't `% === 1e-2 (1/100)`?

Reaktoro/Common/Units.cpp
gdmiron: I think `degR` is standard.

allanleal: You're right - https://en.wikipedia.org/wiki/Rankine_scale

Reaktoro/Common/Units.cpp
gdmiron: you are right, this was added to `%` with `%` inside the string, but as you have shown is not necessary. for percent `%` and permil `%%`.

Reaktoro/Common/Units.cpp
allanleal: Did you remove `%` and `%%`? I can't see them in the code anymore.

allanleal:
Hi @gdmiron . What is the status of this PR? Can we merge? If not, here is a tip: add the word WIP to the title of the PR (e.g., [WIP] Adding more units). This means *work in progress*, and then the PR should not be merged yet. When ready, remove the WIP from the title.

allanleal:
> `u` for micro should be sufficient, if anything I see `mc` sometimes used for micro.

Shall we use just `u` then? In some cases, it looks weird the `mc` (e.g., `mcN`, `mcm`).

Reaktoro/Common/Units.cpp
```diff
@@ -98,11 +98,17 @@ map<string, DerivedUnit> derivedUnitsMap =
 {
     // Nondimensional Unit
     {"1"            , {{1, "1", 1}}},
+    {"percent"      , {{centi, "1", 1}}},
+    {"permil"       , {{milli, "1", 1}}},
+    {"ppt"          , {{milli, "1", 1}}},
```
bjodah: just a drive-by comment, pardon if I'm being intrusive, but I think "ppt" usually denotes "parts per trillion", cf. https://en.wikipedia.org/wiki/Parts-per_notation

allanleal: Hi, not intrusive at all. Thanks a lot. 

> One part per thousand should generally be spelled out in full and not as "ppt" (which is usually understood to represent "parts per trillion")

gdmiron:
If there are no more comments, from my side, we can merge.

PR #61, opened 2019-03-06
New functions to allow setup with species, elements or compounds

armandoeng:
I created new functions, as suggested on the issue.

But, the older functions, should we delete them, put some warning of Deprected, or just leave there?

### DIFF
--- a/Reaktoro/Common/StringList.cpp
+++ b/Reaktoro/Common/StringList.cpp
@@ -109,7 +109,7 @@ StringList::StringList(const std::vector<std::string>& strings)
 {}
 
 StringList::StringList(std::initializer_list<std::string> strings)
-	: _strings(strings.begin(), strings.end())
+: _strings(strings.begin(), strings.end())
 {}
 
 StringList::~StringList()
--- a/Reaktoro/Common/StringList.cpp
+++ b/Reaktoro/Common/StringList.cpp
@@ -108,6 +108,10 @@ StringList::StringList(const std::vector<std::string>& strings)
 : _strings(strings.begin(), strings.end())
 {}
 
+StringList::StringList(std::initializer_list<std::string> strings)
+	: _strings(strings.begin(), strings.end())
+{}
+
 StringList::~StringList()
 {}
 
--- a/Reaktoro/Common/StringList.cpp
+++ b/Reaktoro/Common/StringList.cpp
@@ -122,7 +122,7 @@ auto StringList::strings() const -> const std::vector<std::string>&
 
 StringList::operator const std::vector<std::string>&() const
 {
-	return _strings;
+    return _strings;
 }
 
 } // namespace Reaktoro
--- a/Reaktoro/Common/StringList.cpp
+++ b/Reaktoro/Common/StringList.cpp
@@ -116,5 +116,10 @@ auto StringList::strings() const -> const std::vector<std::string>&
     return _strings;
 }
 
+StringList::operator const std::vector<std::string>&() const
+{
+	return _strings;
+}
+
 } // namespace Reaktoro
 
--- a/Reaktoro/Common/StringList.hpp
+++ b/Reaktoro/Common/StringList.hpp
@@ -63,6 +63,9 @@ public:
     /// Construct a StringList instance with a vector of strings.
     StringList(const std::vector<std::string>& strings);
 
+	/// Construct a StringList with a initializer list
+	StringList(std::initializer_list<std::string> strings);
+
     /// Destroy this StringList instance.
     virtual ~StringList();
 
--- a/Reaktoro/Common/StringList.hpp
+++ b/Reaktoro/Common/StringList.hpp
@@ -69,6 +69,8 @@ public:
     /// Return a vector of strings.
     auto strings() const -> const std::vector<std::string>&;
 
+    operator const std::vector<std::string>&() const;
+
 private:
     std::vector<std::string> _strings;
 };
--- a/Reaktoro/Common/StringList.hpp
+++ b/Reaktoro/Common/StringList.hpp
@@ -63,8 +63,8 @@ public:
     /// Construct a StringList instance with a vector of strings.
     StringList(const std::vector<std::string>& strings);
 
-	/// Construct a StringList with a initializer list
-	StringList(std::initializer_list<std::string> strings);
+    /// Construct a StringList with a initializer list
+    StringList(std::initializer_list<std::string> strings);
 
     /// Destroy this StringList instance.
     virtual ~StringList();
--- a/Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
+++ b/Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
@@ -15,27 +15,30 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this library. If not, see <http://www.gnu.org/licenses/>.
 
-from reaktoro import ChemicalEditor
+from reaktoro import ChemicalEditor, ChemicalSystem
 import pytest
 
-def _CheckChemicalSystems(system_1, system_2):
+def _check_equivalent_chemical_systems(system_1, system_2):
     assert system_1.numElements() == system_2.numElements()
     assert system_1.numSpecies() == system_2.numSpecies()
     assert system_1.numPhases() == system_2.numPhases()
 
 def test_add_phases_right_use():
     """Test the normal use of addAqueousPhase, addGaseousPhase and addMineralPhase."""
-    editor_first_style = ChemicalEditor()
-    editor_first_style.addAqueousPhase("H2O(l) H+ OH- HCO3- CO2(aq) CO3--")
-    editor_first_style.addGaseousPhase("H2O(g) CO2(g)")
-    editor_first_style.addMineralPhase("Graphite")
+    editor1 = ChemicalEditor()
+    editor1.addAqueousPhase("H2O(l) H+ OH- HCO3- CO2(aq) CO3--")
+    editor1.addGaseousPhase("H2O(g) CO2(g)")
+    editor1.addMineralPhase("Graphite")
     
-    editor_second_style = ChemicalEditor()
-    editor_second_style.addAqueousPhase(["H2O(l)", "H+", "OH-", "HCO3-", "CO2(aq)", "CO3--"])
-    editor_second_style.addGaseousPhase(["H2O(g)", "CO2(g)"])
-    editor_second_style.addMineralPhase(["Graphite"])
+    editor2 = ChemicalEditor()
+    editor2.addAqueousPhase(["H2O(l)", "H+", "OH-", "HCO3-", "CO2(aq)", "CO3--"])
+    editor2.addGaseousPhase(["H2O(g)", "CO2(g)"])
+    editor2.addMineralPhase(["Graphite"])
+    
+    system1 = ChemicalSystem(editor1)
+    system2 = ChemicalSystem(editor2)
  
-    _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
+    _check_equivalent_chemical_systems(system1, system2)
     
 def test_add_phases_wrong_use():
     """Test the wrong usage of addAqueousPhase, addGaseousPhase and addMineralPhase."""
@@ -60,17 +63,20 @@ def test_add_phases_wrong_use():
         
 def test_add_phases_with_elements_right_use():
     """Test the normal use of addAqueousPhaseWithElements, addGaseousPhaseWithElements and addMineralPhaseWithElements."""
-    editor_first_style = ChemicalEditor()
-    editor_first_style.addAqueousPhaseWithElements("H C O Ca")
-    editor_first_style.addGaseousPhaseWithElements("H C O")
-    editor_first_style.addMineralPhaseWithElements("Ca C O")
+    editor1 = ChemicalEditor()
+    editor1.addAqueousPhaseWithElements("H C O Ca")
+    editor1.addGaseousPhaseWithElements("H C O")
+    editor1.addMineralPhaseWithElements("Ca C O")
     
-    editor_second_style = ChemicalEditor()
-    editor_second_style.addAqueousPhaseWithElements(["H", "C", "O", "Ca"])
-    editor_second_style.addGaseousPhaseWithElements(["H", "C", "O"])
-    editor_second_style.addMineralPhaseWithElements(["Ca", "C", "O"])
+    editor2 = ChemicalEditor()
+    editor2.addAqueousPhaseWithElements(["H", "C", "O", "Ca"])
+    editor2.addGaseousPhaseWithElements(["H", "C", "O"])
+    editor2.addMineralPhaseWithElements(["Ca", "C", "O"])
     
-    _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
+    system1 = ChemicalSystem(editor1)
+    system2 = ChemicalSystem(editor2)
+    
+    _check_equivalent_chemical_systems(system1, system2)
     
 def test_add_phases_with_elements_wrong_use():
     """Test the wrong usage of addAqueousPhaseWithElements."""
@@ -83,17 +89,20 @@ def test_add_phases_with_elements_wrong_use():
     
 def test_add_phases_with_elements_of_right_use():
     """Test the normal use of addAqueousPhaseWithElementsOf, addGaseousPhaseWithElementsOf and addMineralPhaseWithElementsOf."""
-    editor_first_style = ChemicalEditor()
-    editor_first_style.addAqueousPhaseWithElementsOf("H2O Ca")
-    editor_first_style.addGaseousPhaseWithElementsOf("CO2 H")
-    editor_first_style.addMineralPhaseWithElementsOf("CaCO3")
+    editor1 = ChemicalEditor()
+    editor1.addAqueousPhaseWithElementsOf("H2O Ca")
+    editor1.addGaseousPhaseWithElementsOf("CO2 H")
+    editor1.addMineralPhaseWithElementsOf("CaCO3")
     
-    editor_second_style = ChemicalEditor()
-    editor_second_style.addAqueousPhaseWithElementsOf(["H2O", "Ca"])
-    editor_second_style.addGaseousPhaseWithElementsOf(["CO2", "H"])
-    editor_second_style.addMineralPhaseWithElementsOf(["CaCO3"])
+    editor2 = ChemicalEditor()
+    editor2.addAqueousPhaseWithElementsOf(["H2O", "Ca"])
+    editor2.addGaseousPhaseWithElementsOf(["CO2", "H"])
+    editor2.addMineralPhaseWithElementsOf(["CaCO3"])
     
-    _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
+    system1 = ChemicalSystem(editor1)
+    system2 = ChemicalSystem(editor2)
+    
+    _check_equivalent_chemical_systems(system1, system2)
     
     
     
--- a/Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
+++ b/Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
@@ -73,25 +73,13 @@ def test_add_phases_with_elements_right_use():
     _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
     
 def test_add_phases_with_elements_wrong_use():
-    """Test the wrong usage of addAqueousPhaseWithElements, addGaseousPhaseWithElements and addMineralPhaseWithElements."""
+    """Test the wrong usage of addAqueousPhaseWithElements."""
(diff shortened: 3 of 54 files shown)


Reaktoro/Thermodynamics/Core/ChemicalEditor.cpp
```diff
@@ -572,6 +572,150 @@ auto ChemicalEditor::addMineralPhase(std::string compounds) -> MineralPhase&
...
+auto ChemicalEditor::addAqueousPhaseWithElements(std::initializer_list<std::string> symbols) -> AqueousPhase&
+{
+	return this->addAqueousPhaseWithElements(std::vector<std::string>(symbols.begin(), symbols.end()));
+}
+
+auto ChemicalEditor::addAqueousPhaseWithElements(const std::string& symbols) -> AqueousPhase&
+{
+	const auto elements = split(symbols, " ");
+	return this->addAqueousPhaseWithElements(elements);
+}
```
allanleal: @armandoeng You can use the class `StringList` and then define a single method here. The following method:

~~~c++
auto ChemicalEditor::addAqueousPhaseWithElements(const StringList& symbols) -> AqueousPhase&;
~~~

will be equivalent to defining the following three:

~~~c++
auto ChemicalEditor::addAqueousPhaseWithElements(const std::vector<std::string>& symbols) -> AqueousPhase&;
auto ChemicalEditor::addAqueousPhaseWithElements(std::initializer_list<std::string> symbols) -> AqueousPhase&;
auto ChemicalEditor::addAqueousPhaseWithElements(const std::string& symbols) -> AqueousPhase&;
~~~

The `StringList` class has implicit constructors from `std::string`, `std::vector<std::string>`, and I guess also `std::initializer_list<std::string>` (even if not, the conversion from `std::vector<std::string>` allows the same syntax).

allanleal:
Maybe we want to add the `deprecated` with a message asking the user to use the other methods.
https://en.cppreference.com/w/cpp/language/attributes/deprecated

tadeu:
It would be nice to have tests added together for these functions.

tadeu:
Which methods are going to be deprecated? Could they be removed already?

I think it's ok to make small breaking changes in the API, if it goes in the direction of having a cleaner API. It's important to document how to "migrate" from code that used the old methods, though. If this causes too much hassle to users, then a new version with the "old" (and deprecated) methods could always be added again later.

Reaktoro/Thermodynamics/Core/ChemicalEditor.cpp
armandoeng: Wonderful :)
I thougth that it was just an example when you commented, I didn't think it existed. :(

allanleal:
Alternatively, `addAqueousPhase` could be equivalent to `addAqueousPhaseWithSpecies`. The same for gaseous and mineral phases. 

~~~python
editor = ChemicalEditor()
editor.addAqueousPhase("H2O(l) H+ OH- Na+ Cl- CO2(aq) HCO3- CO3-- Ca++ Mg++")
editor.addGaseousPhase("H2O(g) CO2(g)")
editor.addMineralPhase("Calcite")
editor.addMineralPhase("Dolomite")
~~~

These methods accept a `StringList` object, so that the above could be replaced by:

~~~python
editor = ChemicalEditor()
editor.addAqueousPhase(["H2O(l)", "H+", "OH-", "Na+", "Cl-", "CO2(aq)", "HCO3-", "CO3--", "Ca++", "Mg++"])
editor.addGaseousPhase(["H2O(g)", "CO2(g)"])
editor.addMineralPhase(["Calcite"])
editor.addMineralPhase(["Dolomite"])
~~~

armandoeng:
> 
> 
> Alternatively, `addAqueousPhase` could be equivalent to `addAqueousPhaseWithSpecies`. The same for gaseous and mineral phases.
> 
> ```python
> editor = ChemicalEditor()
> editor.addAqueousPhase("H2O(l) H+ OH- Na+ Cl- CO2(aq) HCO3- CO3-- Ca++ Mg++")
> editor.addGaseousPhase("H2O(g) CO2(g)")
> editor.addMineralPhase("Calcite")
> editor.addMineralPhase("Dolomite")
> ```
> 
> These methods accept a `StringList` object, so that the above could be replaced by:
> 
> ```python
> editor = ChemicalEditor()
> editor.addAqueousPhase(["H2O(l)", "H+", "OH-", "Na+", "Cl-", "CO2(aq)", "HCO3-", "CO3--", "Ca++", "Mg++"])
> editor.addGaseousPhase(["H2O(g)", "CO2(g)"])
> editor.addMineralPhase(["Calcite"])
> editor.addMineralPhase(["Dolomite"])
> ```

Im just in doubt about the documentation.
Example:
```c++
    /// Add a mineral phase in the chemical editor.
    /// This method constructs a MineralPhase object that represents a mineral phase in the system.
    /// The MineralPhase object is created by specifying the names of the species one by one. These
    /// species names must conform to those used in the database that was specified during the
    /// initialization of the ChemicalEditor object, otherwise an exception will be thrown.
    /// The example below describes an usage of this method for the creation of three pure mineral
    /// phases and one solid solution with two mineral species.
    /// ~~~
    /// ChemicalEditor editor;
    ///
    /// // Create a pure mineral phase with only calcite [CaCO3(s)]
    /// editor.addMineralPhaseWithSpecies({"Calcite"});
    ///
    /// // Create a pure mineral phase with only magnesite [MgCO3(s)]
    /// editor.addMineralPhaseWithSpecies("Magnesite");
    ///
    /// // Create a solid solution with mineral species calcite and magnesite
    /// editor.addMineralPhaseWithSpecies({"Calcite", "Magnesite"});
    /// ~~~
    /// An alternative way, in which to prior knowledge of the species names in the database is
    /// needed, consists of specifying a list of chemical element or compounds,
    /// and let the ChemicalEditor to figure out automatically which species from the loaded database
    /// should be added in the phase. This functionality is supported by method
    /// @ref addMineralPhaseWithElements(std::string symbols).
	/// @ref addMineralPhaseWithElementsOf(std::string formulas).
    /// @param species A vector containing the names of the species.
    /// There is a overloaded function that accepts it as a string.
    /// @return A reference to the created MineralPhase object.
    /// @see addAqueousPhaseWithSpecies, addGaseousPhaseWithSpecies
    auto addMineralPhaseWithSpecies(const StringList& species) -> MineralPhase&;

    /// @overload
    auto addMineralPhase(const StringList& species) -> MineralPhase&;
```

And should I get some other doc informing that 
```c++
auto addMineralPhase(std::vector<std::string> species) -> MineralPhase&;
```
was replaced by
```c++
auto addMineralPhase(const StringList& species) -> MineralPhase&;
```

armandoeng:
Maybe we should put a assert to break the program if someone pass a compound or element in auto addMineralPhase(const StringList& species) -> MineralPhase&;

allanleal:
> Maybe we should put a assert to break the program if someone pass a compound or element in auto addMineralPhase(const StringList& species) -> MineralPhase&;

A runtime error will be thrown with a message that there is no such species in the database.

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -200,6 +200,88 @@ class ChemicalEditor
     /// @see addGaseousPhase, addMineralPhase
     auto addAqueousPhase(std::string compounds) -> AqueousPhase&;
```
allanleal: We can delete this method: `auto addAqueousPhase(std::string compounds) -> AqueousPhase&;`

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -200,6 +200,88 @@ class ChemicalEditor
...
+    /// An alternative way, in which to prior knowledge of the species names in the database is
+    /// needed, consists of specifying a list of chemical element, compound, or substance names,
+    /// and let the ChemicalEditor to figure out automatically which species from the loaded database
+    /// should be added in the phase. This functionality is supported by method
+    /// @ref addAqueousPhaseWithElements(const std::vector<std::string>& compounds).
+    /// @param species A vector containing the names of the species.
+    /// There is a overloaded function that accepts as a string.
+    /// @return A reference to the created AqueousPhase object.
+    /// @see addGaseousPhaseWithSpecies, addMineralPhaseWithSpecies
+    auto addAqueousPhaseWithSpecies(const std::vector<std::string>& species) -> AqueousPhase&;
```
allanleal: Both `addAqueousPhase` and `addAqueousPhaseWithSpecies` are equivalent. Not sure if we need this one ending with `WithSpecies`. In which case, we would then only have

~~~c++
addAqueousPhase({ /* list of species names */})
addAqueousPhaseWithElements({ /* list of element symbols */})
addAqueousPhaseWithElementsOf({ /* list of chemical formulas */})
~~~

But I'm OK having both `addAqueousPhase` and `addAqueousPhaseWithSpecies`.

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -200,6 +200,88 @@ class ChemicalEditor
...
+
+    /// @overload
+    auto addAqueousPhaseWithSpecies(std::initializer_list<std::string> species) -> AqueousPhase&;
+    auto addAqueousPhaseWithSpecies(const std::string& species) -> AqueousPhase&;
```
allanleal: We can use `StringList` and cut down the number of overloaded methods.

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -200,6 +200,88 @@ class ChemicalEditor
...
+    /// might only contain one species of interest.
+    /// @param compounds A Vector containing a list of element or compound names.
+    /// There is a overloaded function that accepts as a string.
+    /// @return A reference to the created AqueousPhase object.
+    /// @see addGaseousPhaseWithElements, addMineralPhaseWithElements
+    auto addAqueousPhaseWithElements(const std::vector<std::string>& symbols) -> AqueousPhase&;
+
+    /// @overload
+    auto addAqueousPhaseWithElements(std::initializer_list<std::string> symbols) -> AqueousPhase&;
+    auto addAqueousPhaseWithElements(const std::string& symbols) -> AqueousPhase&;
```
allanleal: The same here - we can reduce these three methods into a single one with `StringList`. The same for the cases below (so I don't get repetitive :) ).

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -332,6 +484,106 @@ class ChemicalEditor
     /// @see addAqueousPhase, addGaseousPhase
     auto addMineralPhase(std::string compounds) -> MineralPhase&;
```
allanleal: We can delete this method.

Reaktoro/Thermodynamics/Core/ChemicalEditor.hpp
```diff
@@ -261,6 +342,77 @@ class ChemicalEditor
     /// @see addAqueousPhase, addMineralPhase
     auto addGaseousPhase(std::string compounds) -> GaseousPhase&;
```
allanleal: We can delete this method.

Reaktoro/Common/StringList.hpp
```diff
@@ -63,6 +63,9 @@ class StringList
     /// Construct a StringList instance with a vector of strings.
     StringList(const std::vector<std::string>& strings);
 
+	/// Construct a StringList with a initializer list
```
tadeu: Seems to be mixing tabs with spaces here

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,95 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+from reaktoro import ChemicalEditor
+import pytest
+
+def _CheckChemicalSystens(system_1, system_2):
```
tadeu: Typo: `Systens` -> `Systems`

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,95 @@
...
+    assert system_1.numElements() == system_2.numElements()
+    assert system_1.numSpecies() == system_2.numSpecies()
+    assert system_1.numPhases() == system_2.numPhases()
+    
+    assert system_1.elements() == system_2.elements()
+    assert system_1.species() == system_2.species()
+    assert system_1.phases() == system_2.phases()
+
+def test_add_phases_rigth_use():
```
tadeu:
```suggestion
def test_add_phases_right_use():
```

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,95 @@
...
+    editor_second_style.addAqueousPhase(["H2O(l)", "H+", "OH-", "HCO3-", "CO2(aq)", "CO3--"])
+    editor_second_style.addGaseousPhase(["H2O(g)", "CO2(g)"])
+    editor_second_style.addMineralPhase(["Graphite"])
+ 
+    _CheckChemicalSystens(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
+    
+def test_add_phases_wrong_use():
+    """Test the wrong usage of addAqueousPhase, addGaseousPhase and addMineralPhase."""
+    editor = ChemicalEditor()
+    with pytest.raises(RuntimeError):
```
tadeu: Should use one `with pytest.raises(RuntimeError):` per line of "wrong usage", otherwise it will only test the first line

Reaktoro/Common/StringList.cpp
```diff
@@ -108,6 +108,10 @@ StringList::StringList(const std::vector<std::string>& strings)
 : _strings(strings.begin(), strings.end())
 {}
 
+StringList::StringList(std::initializer_list<std::string> strings)
+	: _strings(strings.begin(), strings.end())
```
allanleal:
```suggestion
: _strings(strings.begin(), strings.end())
```

Reaktoro/Common/StringList.cpp
```diff
@@ -116,5 +120,10 @@ auto StringList::strings() const -> const std::vector<std::string>&
     return _strings;
 }
 
+StringList::operator const std::vector<std::string>&() const
+{
+	return _strings;
```
allanleal:
```suggestion
    return _strings;
```

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,99 @@
...
+    editor_first_style.addAqueousPhase("H2O(l) H+ OH- HCO3- CO2(aq) CO3--")
+    editor_first_style.addGaseousPhase("H2O(g) CO2(g)")
+    editor_first_style.addMineralPhase("Graphite")
+    
+    editor_second_style = ChemicalEditor()
+    editor_second_style.addAqueousPhase(["H2O(l)", "H+", "OH-", "HCO3-", "CO2(aq)", "CO3--"])
+    editor_second_style.addGaseousPhase(["H2O(g)", "CO2(g)"])
+    editor_second_style.addMineralPhase(["Graphite"])
+ 
+    _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
```
allanleal:
```suggestion
    system1 = ChemicalSystem(editor_first_style)
    system2 = ChemicalSystem(editor_second_style)
    _CheckChemicalSystems(system1, system2)
```

allanleal: What about we simplify the variable names to `editor1` and `editor2` instead? ;)

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,99 @@
...
+    editor_first_style.addAqueousPhaseWithElements("H C O Ca")
+    editor_first_style.addGaseousPhaseWithElements("H C O")
+    editor_first_style.addMineralPhaseWithElements("Ca C O")
+    
+    editor_second_style = ChemicalEditor()
+    editor_second_style.addAqueousPhaseWithElements(["H", "C", "O", "Ca"])
+    editor_second_style.addGaseousPhaseWithElements(["H", "C", "O"])
+    editor_second_style.addMineralPhaseWithElements(["Ca", "C", "O"])
+    
+    _CheckChemicalSystems(editor_first_style.createChemicalSystem(), editor_second_style.createChemicalSystem())
```
allanleal: Same changes here and possibly below too (as shown above).

Reaktoro/Thermodynamics/Core/ChemicalEditor.cpp
```diff
@@ -47,10 +48,10 @@
 namespace Reaktoro {
 namespace {
 
-auto collectElementsInCompounds(std::vector<std::string> compounds) -> std::vector<std::string>
+auto collectElementsInCompounds(const std::vector<std::string>& compounds) -> std::vector<std::string>
```
allanleal: Thanks to the `const&` - sometimes I'm bothered with how long the argument list gets and just ignore them (especially when we can use `std::move` operator)! ;)

demos/cpp/demo-activity-debye-huckel.cpp
```diff
@@ -25,7 +25,7 @@ int main()
     db.setPHREEQC();
 
     ChemicalEditor editor;
-    editor.addAqueousPhase("H2O NaCl CaCl2 CO2")
+    editor.addAqueousPhaseWithElementsOf("H2O NaCl CaCl2 CO2")
```
allanleal: This looks much more self explanatory now!

demos/cpp/demo-chemicalsolver-equilibrium.cpp
```diff
@@ -23,7 +23,7 @@
 //    Index npoints = 10;
 //
 //    ChemicalEditor editor;
-//    editor.addAqueousPhase("H O Na Cl C Ca Mg");
+//    editor.addAqueousPhaseWithElements("H O Na Cl C Ca Mg");
```
allanleal: Very nice!

tests/test_reaktoro_misc.py
```diff
@@ -26,11 +26,11 @@ def verify_dict(chemical):
     assert dict1['teste'] == dict2['teste'] and dict1[chemical] == dict2[chemical]
```
allanleal:
```suggestion
    assert dict1['test'] == dict2['test'] and dict1[chemical] == dict2[chemical]
```

Reaktoro/Thermodynamics/Core/ChemicalEditor-test.py
```diff
@@ -0,0 +1,99 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+from reaktoro import ChemicalEditor
+import pytest
+
+def _CheckChemicalSystems(system_1, system_2):
```
allanleal:
```suggestion
def _check_equivalent_chemical_systems(system1, system2):
```

allanleal: Need to change occurrences in the rest of the file.

Reaktoro/Thermodynamics/Core/ChemicalEditor.cpp
armandoeng: Yeah, I did the same here on souring. But Tadeu get to my attention. :P

Reaktoro/Common/StringList.cpp
armandoeng: Sorry for all these format problems. Normally I don't get to them and let the hook get them. 
The problem is with my visual. It puts tabs. I fix it, but everytime I rebuild the solution with ninja, it come back to tab default.

tests/test_reaktoro_misc.py
```diff
@@ -10,27 +10,27 @@ def create_minimum_system():
...
+    assert dict1['test'] == dict2['test'] and dict1[chemical] == dict2[chemical]
 
 
-@pytest.mark.xfail
 def test_chemical_state_with_empty_system():
     system = ChemicalSystem()
     state = ChemicalState(system)
-    repr(state)
+    with pytest.raises(RuntimeError):        
+        repr(state)
```
tadeu: Hmm, it's not very good to have `repr` raising exceptions.. it's used in places that don't expect exceptions (e.g., for debugging, etc.). Could `repr` just return something that says that the system is empty?

allanleal: This was fixed today - we should remove this test, because it is already tested in `ChemicalState-test.py`. The failure above used to happen because an aqueous phase was always presumed in the system (and thus, at some point, `indexPhase` method would return an out of bound index!).

armandoeng: removed

PR #63, opened 2019-03-11
Fix bindings for ChemicalProperties::subvolume

tadeu:
Fixes missing `#include <pybind11/stl.h>`.

This also fixes some compilations, the fact that the include is missing from one .cpp also affects bindings from other .cpp files, such as `ChemicalState::setSpeciesAmounts` (which receives an `Indices` vector).


tadeu:
I'm merging this since it's a quick fix, but feel free to comment, @allanleal, I'll apply suggestions later

allanleal:
> I'm merging this since it's a quick fix, but feel free to comment, @allanleal, I'll apply suggestions later

Of course @tadeu ! Thanks for this.

PR #64, opened 2019-03-11
Change to always include required pybind11 headers

tadeu:
Picking individual headers from `pybind11` is errorprone, because, if something is missing, it will present errors only at runtime. If we are lucky it will present a nice message asking if we forgot to include something, but sometimes it will fail with hard-to-track problems.

I've noticed that missing an include like `#include <pybind11/stl.h>` in one compilation unit might affect bindings defined in other compilation units (it depends on the order of linkage, sometimes it might happen, sometimes not, and is a very tricky problem to find out).

For a small increment in compilation time, I think it's worthy to have this increased "safety".

### DIFF
--- a/python/PyReaktoro/Common/PyAutoDiff.cpp
+++ b/python/PyReaktoro/Common/PyAutoDiff.cpp
@@ -15,10 +15,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this library. If not, see <http://www.gnu.org/licenses/>.
 
-// pybind11 includes
-#include <pybind11/pybind11.h>
-#include <pybind11/eigen.h>
-namespace py = pybind11;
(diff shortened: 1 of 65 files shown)


allanleal:
Hi @tadeu . Thanks for this. Just a small concern: did the compilation time increased substantially by including more pybind11 files (in some cases, some of them will not be needed, e.g., `stl.h` if no stl container being used in the C++ interface). If not, you're good to go.

allanleal:
> For a small increment in compilation time, I think it's worthy to have this increased "safety".

OK, missed this. So, the increase in compilation time was not so much? Because of our caching mechanisms (local, travis, and appveyor) I think this should be ok.

tadeu:
Those builds here are between 12 and 14 minutes, so if there's any change them seem unnoticeable :)

Issue #65, opened 2019-03-12
Error when trying to calculate equilibrium with just a gaseous phase


martinvoigt:
I wasn't able to calculate the equilibrium state of systems where just a gaseous phase would be present:

```Python
db = Database('supcrt98')
editor = ChemicalEditor(db)
editor.addGaseousPhase('H O')

system = ChemicalSystem(editor)

problem = EquilibriumProblem(system)
problem.setTemperature(150, 'celsius')
problem.setPressure(1,'bar')
problem.add('H2O', 1.0, 'kg')

state = equilibrate(problem)

print(state)
```

This is the error I got:
```
RuntimeError: 
****************************************************************************************
*** Error: Could not get a reference to a Phase instance with given index.
*** Reason: The given index 1 is out of bounds.
*** Location:  This error was encountered in :236.
****************************************************************************************
```

Seems there is no phase present at all... Playing around, sometimes I already got an error at `equilibrate(problem)`, telling me no equilibrium was found - but I couldn't reproduce this one again unfortunately.

I tried finding out more by trying to see what `state.properties().phaseAmounts()` returns, but wasn't able to get much out of it, as I didn't find much documentation on the `ChemicalVector` type yet.

It works if an aqueous phase is added using `editor.addAqueousPhase('H O')` to the problem though, so not sure if this is just a problem with displaying the output or with equilibrium calculations themselves.

Apart from this issue: Would it be possible to add an option to set the pressure to the vapor pressure at the specific temperature/composition when defining a problem? I guess this could be done by a user by running some iterations to find that phase boundary, but having it built in would be convenient I think...
Maybe that's already there, but I didn't find it in the docs yet.
I didn't think this through all the way, but this might require setting the volume to a fixed value (instead of the pressure) in such a problem. Is that possible, or will it ever be?

Generalizing this: How about adding some inverse-solving capabilities? An obvious interesting thing would be finding phase boundaries (such as water-vapor, as described above, but also between minerals etc.).

Another related "gimmick" would be to be able to define a certain pH for an equilibrium problem, which is realized by adding the right amount of an acid or base (such as HCl or NaOH). In PHREEQC, this can be done by defining a custom phase `H+` with an equilibrium constant of 0, and then forcing the saturation index of this phase to `-pH` using HCl, NaOH, or other chemicals. Being able to set saturation indices to a specific value (not only 0) is a useful capabilities in general as well...

allanleal:
[...]

PR #66, opened 2019-03-13
Fix chemical state output

allanleal:
This PR is to address issue #65 - printing chemical state without aqueous phase fails.

This is just a quick solution. The plan is to remove this code in `ChemicalState` related to aqueous concepts and move it to another class dedicated for aqueous phase properties and states.


Reaktoro/Core/ChemicalState.cpp
```diff
@@ -903,20 +895,32 @@ auto operator<<(std::ostream& out, const ChemicalState& state) -> std::ostream&
...
-    out << std::left << std::setw(25) << "Alkalinity [eq/L]";
-    out << std::endl << bar2 << std::endl;
-    out << std::left << std::setw(25) << I;
-    out << std::left << std::setw(25) << pH;
-    out << std::left << std::setw(25) << pE;
-    out << std::left << std::setw(25) << Eh;
-    out << std::left << std::setw(25) << alk;
-    out << std::endl << bar1 << std::endl;
+    // Check if there is an aqueous phase before printing aqueous states
+    if(system.indexPhase("Aqueous") < system.numPhases())
```
tadeu: It seems tricky that `indexPhase("Aqueous")` might return an invalid index, and also tricky because this relies on the assumption that the index for the Aqueous phase is the last one.

A change that could be considered later: `indexPhase("phase")` would throw an exception if the phase is not present, and `hasPhase("phase")` could be added to check for phase presence so that `try/catch` isn't required just for this.

allanleal: > A change that could be considered later: indexPhase("phase") would throw an exception if the phase is not present

There is a method called `indexPhaseWithError` that does that! ;)

allanleal: >  and hasPhase("phase") could be added to check for phase presence

This is nice. The thing is that internally, this will be implemented by determining the index of the species of interest (and the index later discarded). Then the user will most likely later call `indexPhase` again to get the index. The check `index < size` is a C++ usual thing. In python, the index method of a list raises an exception if the item is not found.

tadeu: Ah, good :)

In this case, for safety, perhaps it could be like this: `indexPhaseWithError` renamed to `indexPhase` (public), current `indexPhase` renamed to `indexPhaseWithoutCheck` and made protected or private (if it makes sense to keep having it for performance reasons).

allanleal: > In this case, for safety, perhaps it could be like this: indexPhaseWithError renamed to indexPhase (public), current indexPhase renamed to indexPhaseWithoutCheck and made protected or private (if it makes sense to keep having it for performance reasons).

I think here we might want to give users the two possibilities, like `std::vector::at` and `std::vector::operator[]`, where the former has bound check for safety, and the latter does not for efficiency.

The `indexPhase` raising a runtime error and `indexPhaseWithoutCheck` is perhaps indeed more adequate than the current understanding (`indexPhase` has not check).

allanleal:
Thanks @tadeu !

Reaktoro/Core/ChemicalState.cpp
allanleal: Just merged, but we can continue this discussion (here or in another issue)! ;)

tadeu: Yeah, I understand the concerns about the double-search that is needed for the `has`/`find` sequence... this is why a `std::find` will return an invalid iterator pointing to `end` if it hasn't been found. 

Usually I tend to give more weight to the safety of the user (minimize segfaults probability, etc.) than to performance "everywhere" (of course, performance at possible bottlenecks are still very important).

In this case, there is another possibility: `indexPhase` could return a `std::optional<Index>`, so it would contain nothing if the phase doesn't exist, and would only perform one search.

allanleal: The optional idea is interesting. With C++17, one could do (using `if` with an initialize):

~~~c++
if (auto idx = system.indexPhase("Aqueous"); idx)
{
    // do stuff for aqueous phase
}
~~~

PR #75, opened 2019-04-11
Add test to locale problem on reaktoro

armandoeng:
change function elements to be const

### DIFF
--- a/Reaktoro/Thermodynamics/Core/Database-test.py
+++ b/Reaktoro/Thermodynamics/Core/Database-test.py
@@ -19,38 +19,63 @@ from reaktoro import Database
 import pytest
 
 import locale
-import subprocess
 import platform
 
-def treat_locales_by_platform(locales):
-    system = platform.system()
-    if system == 'Java':
-        return []
-    elif system == 'Linux':
-        return locales
-    locales = [locale for locale in locales if not 'utf8' in locale and not '@' in locale and not 'POSIX' in locale]
-    locales = [locale.replace('_', '-') for locale in locales]
-    return locales
+LANGUAGES = {
+    'bg_BG': 'Bulgarian',
+    'cs_CZ': 'Czech',
+    'da_DK': 'Danish',
+    'de_DE': 'German',
+    'el_GR': 'Greek',
+    'en_US': 'English',
+    'es_ES': 'Spanish',
+    'et_EE': 'Estonian',
+    'fi_FI': 'Finnish',
+    'fr_FR': 'French',
+    'hr_HR': 'Croatian',
+    'hu_HU': 'Hungarian',
+    'it_IT': 'Italian',
+    'lt_LT': 'Lithuanian',
+    'lv_LV': 'Latvian',
+    'nl_NL': 'Dutch',
+    'no_NO': 'Norwegian',
+    'pl_PL': 'Polish',
+    'pt_PT': 'Portuguese',
+    'ro_RO': 'Romanian',
+    'ru_RU': 'Russian',
+    'sk_SK': 'Slovak',
+    'sl_SI': 'Slovenian',
+    'sv_SE': 'Swedish',
+    'tr_TR': 'Turkish',
+    'zh_CN': 'Chinese',
+}
 
-def find_locales():
-    out = subprocess.run(['locale', '-a'], stdout=subprocess.PIPE).stdout
+def is_locale_working(loc):
     try:
-        res = out.decode('utf-8')
+        locale.setlocale(locale.LC_NUMERIC, loc)
+        return True
     except:
-        res = out.decode('latin-1')
-    return res.rstrip('\n').splitlines()
+        return False
 
-def set_locale():
-    available_locales = treat_locales_by_platform(find_locales())
-    for loc in available_locales:
-        try:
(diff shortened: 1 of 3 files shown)


armandoeng:
@allanleal , I did a quick fix on database, which changes locale to C, and return it to the old locale, so avoids this problem. Also, I added some tests to verify it is working, even for different locales.

Reaktoro/Thermodynamics/Core/Database.cpp
```diff
@@ -393,11 +393,13 @@ struct Database::Impl
...
     MineralSpeciesMap mineral_species_map;
 
-    Impl()
-    {}
+    Impl() = default;
 
     Impl(std::string filename)
     {
+        const std::string old_locale(std::setlocale(LC_NUMERIC, nullptr));
+        std::setlocale(LC_NUMERIC, "C");
```
tadeu: It needs an `#include <clocale>` at the top

tadeu: If `parse` throws an exception, the original `locale` won't be restored. A "guard" object may be created for this (or it could also use something like `gsl::finally` if it depended on `GSL`):

```cpp
class change_locale_to_default_in_scope
{
public:
    change_locale_to_default_in_scope() : old_locale(std::setlocale(LC_NUMERIC, nullptr)) {}
    ~change_locale_to_default_in_scope() { std::setlocale(LC_NUMERIC, this->old_locale.c_str()); }
private:
    const std::string old_locale;
};
```

then just declare it at the beggining of the method:

```
    Impl(std::string filename)
    {
        auto _locale_guard = change_locale_to_default_in_scope();
```

Reaktoro/Thermodynamics/Core/Database.cpp
```diff
@@ -458,9 +462,18 @@ struct Database::Impl
         mineral_species_map.insert({species.name(), species});
     }
 
-    auto elements() -> std::vector<Element>
+    auto elements() const-> std::vector<Element>
     {
-        return collectValues(element_map);
+        std::vector<Element> elements{};
```
tadeu: Why did this change?

armandoeng: I wanted to make elements() be a const function. As it was, if you changed anything at the elements of the returned vector, that would alter the elements inside the database. 
This happens because Element works as a shared pointer to the attributes of Element, and every copy is a shallow copy.

armandoeng: I wanted to make elements() be a const function. As it was, if you changed anything at the elements of the returned vector, that would alter the elements inside the database. 
This happens because Element works as a shared pointer to the attributes of Element, and every copy is a shallow copy.

armandoeng: I wanted to make elements() be a const function. As it was, if you changed anything at the elements of the returned vector, that would alter the elements inside the database. 
This happens because Element works as a shared pointer to the attributes of Element, and every copy is a shallow copy.

tadeu:
AppVeyor is failing (something to do with unsupported locale)

armandoeng:
> AppVeyor is failing (something to do with unsupported locale)

Yes, it changes from computer to computer. The function findlocales was supposed to find all supported locales on the computer, but looks like it didn't worked as planned. 
Put some try to avoid breaking test. Trying to find a better solution.

armandoeng:
@alexandrebbruno, you will have the task to clone this branch and test Database-test.py on Mac, ok? ;)

Reaktoro/Thermodynamics/Core/Database-test.py
```diff
@@ -0,0 +1,125 @@
...
+    assert pytest.approx(0.040078) == molar_mass_map['Ca']
+    assert pytest.approx(0.0159994) == molar_mass_map['O']
+    assert pytest.approx(0.03206499999999) == molar_mass_map['S']
+    assert pytest.approx(0.02298977) == molar_mass_map['Na']
+    assert pytest.approx(0.035453) == molar_mass_map['Cl']
+    assert pytest.approx(0.08762) == molar_mass_map['Sr']
+
+
+def test_locale_problem_with_pugixml():
+    old_locale = locale.setlocale(locale.LC_NUMERIC)
```
tadeu: Just one more suggestion :)

Create a fixture that will keep the old_locale and restore it in the end, like:

```python
@pytest.fixture
def restore_locale():
    old_locale = locale.setlocale(locale.LC_NUMERIC)
    yield
    locale.setlocale(locale.LC_NUMERIC, old_locale)
```

it can be used like this:

```python
@pytest.mark.usefixtures("restore_locale")
def test_locale_problem_with_pugixml():
    ...
```

this guarantees that the test won't leak a changed locale to other tests (it could happen if, for example, an exception is raised before line 103 here)

armandoeng: One new thing I learned about pytest. Thnks. :)

tadeu:
> Good Thinking. I will ask to @alexandrebbruno to test this on Mac before we merge.

No worries, we can fix it later if it doesn't pass on macOS

PR #77, opened 2019-04-12
Improve CMake so that it won't fail if REAKTORO_PYTHON_INSTALL_PREFIX…

tadeu:
… is passed as a string and with backslashes.

This is a very minor change, to make it more robust to a problem that I stumbled upon in a conda-forge PR: https://github.com/conda-forge/reaktoro-feedstock/pull/6


tadeu:
Thanks, and ah, no worries about delaying, one day may be relatively quick, and we know you are very busy ☺️

PR #82, opened 2019-05-27
Cpp reactivetransportsolver demo

mtsveta:
The demo includes profiling of the ReactiveTransportSolver with Convectional and Smart Equilibrium solvers


Reaktoro/Transport/TransportSolver.hpp
```diff
@@ -204,7 +200,7 @@ class Mesh
 };
 
 /// A class for solving advection-diffusion problem.
-/// Eq: du/dt + v*du/dx = D*d�u/dx�
+/// Eq: du/dt + v*du/dx = D*d�u/dx�
```
allanleal:
```suggestion
/// Eq: du/dt + v*du/dx = D*d²u/dx²
```
You may want to change your text editor to UTF-8 encoding.

Reaktoro/Transport/TransportSolver.hpp
```diff
@@ -239,6 +235,10 @@ class TransportSolver
     /// Setup coefficient matrix of the diffusion problem and factorize.
     auto initialize() -> void;
 
+    /// Initialize the transport solver before method @ref step is executed.
+    /// Setup coefficient matrix of the diffusion-advection problem and factorize.
+    auto initializeFullImplicit() -> void;
```
allanleal: Do we need to initialize methods? Can we choose only one here or merge their features if possible?

Reaktoro/Transport/TransportSolver.hpp
```diff
@@ -247,6 +247,14 @@ class TransportSolver
     /// @param q The source rates vector ([same unit considered for u]/m)
     auto step(VectorRef u, VectorConstRef q) -> void;
 
+    /// Step the transport solver.
+    /// This method solve one step of the transport solver equation, using an explicit approach for
+    /// advection and total implicit for diffusion. The amount resulted from the advection it is
+    /// passed to diffusion problem as a "source".
+    /// @param[in,out] u The solution vector
+    /// @param q The source rates vector ([same unit considered for u]/m)
+    auto step_implicit_fvm(VectorRef u, VectorConstRef q) -> void;
```
allanleal: Also here. We should have only one method. If this method should operate differently, we should provide this via an `options` argument.

demos/cpp/demo-reactivetransportsolver-calcite-brine.cpp
```diff
@@ -0,0 +1,280 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+// C++ includes
+#include <algorithm>    // for using transform
+#include <sstream>      // for using stringstream
+#include <iomanip>      // for setprecition
+#include <tuple>
+
+#if defined _WIN32      // for creating a new folder
```
allanleal: We should use the C++17 `std::filesystem` library:
https://en.cppreference.com/w/cpp/filesystem

To avoid these codes specific for Windows, for example.

demos/python/demo-reactivetransportsolver-calcite-brine.py
```diff
@@ -28,18 +35,127 @@
 
 # Step 3: Define parameters for the reactive transport simulation
 xl = 0.0          # the x-coordinate of the left boundary
-xr = 100.0        # the x-coordinate of the right boundary
-nsteps = 100      # the number of steps in the reactive transport simulation
-ncells = 100      # the number of cells in the discretization
+xr = 0.1          # the x-coordinate of the right boundary
+ncells = 10      # the number of cells in the discretization
```
allanleal: I think 10 is a very small number of cells. We should have 50 or 100 here. But then we need to make sure the smaller cell sizes do not compromise the stability of the demo file.

demos/python/demo-reactivetransportsolver-calcite-brine.py
```diff
@@ -28,18 +35,127 @@
...
+#------------------------------------------------------------------------------#
+# Reactive transport simulations
+#------------------------------------------------------------------------------#
+
 # Step 4: Construct the chemical system with its phases and species
-editor = ChemicalEditor()
-editor.addAqueousPhaseWithElementsOf('H2O NaCl CaCl2 MgCl2 CO2')
+db = Database('supcrt98.xml')
+editor = ChemicalEditor(db)
+#editor.addAqueousPhase(['H2O(l)', 'H+', 'OH-', 'Na+', 'Cl-', 'Ca++', 'Mg++', 'HCO3-', 'CO2(aq)', 'CO3--'])
```
allanleal:
```suggestion
```

demos/python/demo-reactivetransportsolver-calcite-brine.py
```diff
@@ -28,18 +35,127 @@
...
+editor.addAqueousPhase('H2O(l) H+ OH- Na+ Cl- Ca++ Mg++ HCO3- CO2(aq) CO3--')
+#editor.addAqueousPhaseWithElementsOf('H2O NaCl CaCl2 MgCl2 CO2')
```
allanleal:
```suggestion
```

Reaktoro/Transport/TransportSolver.cpp
```diff
@@ -246,29 +246,91 @@ TransportSolver::TransportSolver()
...
+        const double c = -alpha;
         A.row(icell) << a, b, c;
     }
 
     // Assemble the coefficient matrix A for the boundary cells
-    A.row(icell0) << 0.0, 1.0 + 4.5*beta, -1.5*beta; // prescribed amount on the wall and approximatin deriveted by forward diference approximation with second order error
-    A.row(icelln) << -beta, 1.0 + beta, 0.0;   // du/dx = 0 at the right boundary
+    // Our derivation
+    // A.row(icell0) << 0.0, 1.0 + alpha, -alpha;            // forward difference approximation with first order error
+    // A.row(icelln) << 0.0, 1.0, 0.0;                       // d/dx = 0 (zero flux) at the right boundary
```
allanleal: Are we going to assume  d/dx = 0 or  d/dx (D du/dx) = 0 on the last cell? The latter being equivalent to the incoming diffusion flux is the same as the outgoing.

Reaktoro/Transport/ReactiveTransportSolver.hpp
```diff
@@ -0,0 +1,201 @@
...
+    EquilibriumProfiler(Profiling what);
+
+    auto updateLearning(int step) -> void;
+    auto updateEstimating(int step, std::vector<double> time_partition) -> void;
+
+    auto consoleOutput(int step) -> void;
+    auto fileOutput(const std::string & file)-> void;
+
+    /// The vector of the elapsed CPU times for learning and estimating time
+    std::vector<double> learn_times;
```
allanleal: Prefer a struct with proper field names instead of relying on indices and `std::vector`.

Reaktoro/Transport/ReactiveTransportSolver.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    auto output() -> ChemicalOutput;
+
+    auto initialize() -> void;
+
+    auto step(ChemicalField& field) -> void;
+
+    auto step_tracked(ChemicalField& field) -> void;
+
+    auto profile(Profiling what) -> Profiler;
```
allanleal: I think we should avoid this tight dependency of class `ReactiveTransportSolver` with class `Profiling`. Here is my suggestion.

We define a `ReactiveTransportResult` struct, and change:

`auto step(ChemicalField& field) -> void;`

to 

`auto step(ChemicalField& field) -> ReactiveTransportResult;`

We then make sure that `ReactiveTransportResult` contains sufficient information collected during the reactive transport step operation. 

This can later be used by other classes that do analysis of the algorithm.

~~~c++
ReactiveTransportSolver transport(/*...*/);
ReactiveTransportResult result;
ReactiveTransportProfiler profiler;

...

// In a for loop
result = transport.step(field);
profiler.process(result);
profiler.output("some-file-name.txt");
~~~

Reaktoro/Equilibrium/SmartEquilibriumSolver.cpp
```diff
@@ -194,12 +202,98 @@ struct SmartEquilibriumSolver::Impl
         return res;
     }
 
+    /// Estimate equilibrium state using sensitivity derivatives
+    auto estimate_tracked(ChemicalState& state, double T, double P, VectorConstRef be) -> EquilibriumResult
```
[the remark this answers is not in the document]
mtsveta: Yes. I just wanted to separate the methods with profiling and without.

Reaktoro/Equilibrium/SmartEquilibriumSolver.hpp
```diff
@@ -68,10 +68,14 @@ class SmartEquilibriumSolver
     /// Estimate the equilibrium state using sensitivity derivatives.
     auto estimate(ChemicalState& state, double T, double P, VectorConstRef be) -> EquilibriumResult;
 
+    /// Estimate the equilibrium state using sensitivity derivatives.
+    auto estimate_tracked(ChemicalState& state, double T, double P, VectorConstRef be) -> EquilibriumResult;
```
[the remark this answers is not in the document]
mtsveta: Yes, sounds good!

Reaktoro/Transport/TransportSolver.hpp
mtsveta: Sure. I will add simple `options` definition for now, which distinguishes between just finite volume schemes. Later we can extend it.

Reaktoro/Transport/TransportSolver.cpp
mtsveta: No, we are planning to assume free boundary on the right side, as it was done here: 
```
A.row(icelln) << - beta, 1.0 + beta, 0.0;           // right boundary (free)
```
The line with ` A.row(icelln) << -beta, 1.0 + beta, 0.0;   // du/dx = 0 at the right boundary` is an old one. I need to still text and fix that scheme

PR #86, opened 2019-07-02
Update pybind11 to 2.3.0 and Boost to 1.70

tadeu:
* Remove pybind11 code from internal `thirdparty` libraries


cmake/ReaktoroFindDeps.cmake
```diff
@@ -1,21 +1,13 @@
 # Find Boost library
 find_package(Boost REQUIRED)
 
-# Find pybind11 library (if needed) https://github.com/pybind/pybind11
+# Find pybind11 library (if needed)
 if(REAKTORO_BUILD_PYTHON)
-
-    if(NOT PYBIND11_PYTHON_VERSION AND NOT PYTHON_EXECUTABLE)
-        find_package(PythonInterp REQUIRED) # PYTHON_EXECUTABLE gets set here
-    endif()
```
allanleal: I removed this old logic here (prior to the use of conda as a dependency manager), where we may end up finding PythonInterp (which is actually deprecated by cmake).

allanleal:
@tadeu , I'' merge this now, and try to enable CI/conda-forge for macOS after this merge.

PR #88, opened 2019-07-03
Enable Travis CI for macOS

allanleal:
This PR enables CI for Reaktoro (Travis) in macOS.


allanleal:
@tadeu @alexandrebbruno 

Hi guys. Here is what I observe from the tests that are failing in macOS (check the [build log here](https://travis-ci.org/reaktoro/reaktoro/jobs/553629418#L47), everything compiles well; it's just some tests that are failing, but I think they should not actually be marked failed).

So, you'll see that there are some coarse mismatches sometimes for the `Element Potentials`:

~~~log
E       Element Dual Potential [kJ/mol]:
E          obtained_Element Dual Potential [kJ/mol]  expected_Element Dual Potential [kJ/mol]                 diff
E       0                    -410.46190778516523778                    -410.46269082044562992  0.00078303528039214
E       1                     -42.57934469140845124                     -42.60736660213529348  0.02802191072684224
E       2                    -125.70965423031513808                    -125.70984998913513664  0.00019575881999856
E       3                    -343.37038747421598828                    -343.34236556348912472  0.02802191072686355
E       4                      12.16882983544884667                      12.16922135308898056  0.00039151764013390
E       5                      78.79901613579366426                      78.77103514454873334  0.02798099124493092
~~~

When the formula matrix *A* has linearly dependent rows (it's not full-rank), there is a linear combination of element potentials that are acceptable. Right now, I can't fix the solver to ensure that exactly the same solution for the element potentials is found for all OSs; apparently this happens for Linux and Windows already, maybe by chance, but not for macOS.

I would then propose that we do not test these element potentials (at least for now). We instead test the species chemical potentials (`mu`) and since `mu = transpose(A)*y`, where `y` is the vector of element potentials.

I observe, however, that there are also some mismatches in the species chemical potentials:

~~~log
E       Potential [kJ/mol]:
E           obtained_Potential [kJ/mol]  expected_Potential [kJ/mol]                  diff
E       2        -658.41636363416648692       -658.40703800273649904   0.00932563142998788
E       3        -228.55979800844053784       -230.84966246227443776   2.28986445383389992
E       4        -175.00105785560643312       -175.00104622885353933   0.00001162675289379
E       5        -123.63156881178342417       -123.63155221872800382   0.00001659305542034
E       6        -149.41937408581731006       -149.41933668812814062   0.00003739768916944
E       7        -150.01355707471270762       -150.01349718410108380   0.00005989061162381
E       8        -135.98838872574273751       -135.98973275813995087   0.00134403239721337
E       11       -275.62122878045710195       -275.62121794457726764   0.00001083587983430
E       12       -717.14209699569516943       -678.87364343026729330  38.26845356542787613
E       13       -263.62529612967330195       -263.62681425172627314   0.00151812205297119
E       14       -221.06146975817318889       -221.06239604869256254   0.00092629051937365
E       15       -138.32485976956363061       -138.32534843640664235   0.00048866684301174
E       16       -205.86027099099075599       -205.85963028019557441   0.00064071079518158
E       19       -554.89795380194857444       -554.89634539260089241   0.00160840934768203
E       20        -92.22811076846659262        -84.40772778722251246   7.82038298124408016
E       21       -292.15986419022175369       -292.15819129243038788   0.00167289779136581
~~~

These are most often minor though, and only more major diffs for a few species. My guess is that these are species with tiny amounts, which we should ignore during the test.

A side thing. Have a look at these lines from the log:

~~~log
tests/conftest.py:244: AssertionError
_ test_equilibrate_overload_2[Eq Prob-H2O CO2 NaCl and Halite already dissolved at 60 \xb0C and 300 bar] 
~~~

Would you guys agree to instead of using `ids` such as:

 `"Eq Prob-H2O CO2 NaCl and Halite already dissolved at 60 °C and 300 bar"` 

we use the name of the function:

`equilibrium_problem_with_h2o_co2_nacl_halite_dissolved_60C_300bar`

in the `@pytest.mark.parametrize` statements? I believe this is easier to jump to the point in the code where the fixture is created. But maybe there is a simple way figuring out this and I don't know.

tadeu:
> These are most often minor though, and only more major diffs for a few species. My guess is that these are species with tiny amounts, which we should ignore during the test.

Perhaps we could use Gibbs energy



instead of chemical potential, and then define an absolute tolerance. This way, the relative tolerance would be higher for species that are present in very small quantities. The problem with this approach is that some code changes that might cause differences in results (but shouldn't) might slip under the radar (tests wouldn't fail).

Another solution would be to have different expected results for different platforms, but this is usually very troublesome to maintain.

> Would you guys agree to instead of using ids such as:
> `"Eq Prob-H2O CO2 NaCl and Halite already dissolved at 60 °C and 300 bar"`
> we use the name of the function:
> `equilibrium_problem_with_h2o_co2_nacl_halite_dissolved_60C_300bar`

I agree!

tadeu:
I've merged `master` here, just to check if the changes from #89 would be helpful (they fixed some problem of different values on our side), but unfortunately they didn't make a change

allanleal:
@tadeu @alexandrebbruno 

We are almost there here to support macOS build. Note I've merged changes from PR #93 here.

Could you guys take care of this issue, please:

~~~
0.07s$ python ci/check_compiled_files.py
Traceback (most recent call last):
  File "ci/check_compiled_files.py", line 13, in <module>
    assert (artifacts / 'lib/Reaktoro.lib').is_file()
AssertionError
The command "python ci/check_compiled_files.py" exited with 1.
~~~

I think once this is resolved, we are done and sorted out this issue.

tadeu:
Thanks a lot @allanleal, it's awesome that the different values problem in macOS is solved now :)

PR #89, opened 2019-07-03
Fix some different values for different platforms

tadeu:
I'm trying to reproduce an issue we have with different values in different platforms.


Reaktoro/Core/ChemicalState.cpp
```diff
@@ -822,7 +822,7 @@ auto operator<<(std::ostream& out, const ChemicalState& state) -> std::ostream&
 
     // Set output in scientific notation
     auto flags = out.flags();
-    out << std::setprecision(6);
+    out << std::setprecision(17);
```
tadeu: This is temporary (I was using print's to pinpoint differences, so it was helpful), although I thought of adding a `set_precision` configuration to `ChemicalState`, so that it's configurable at runtime.

Reaktoro/Math/LU.cpp
```diff
@@ -137,7 +137,7 @@ auto LU::trsolve(MatrixConstRef B) -> Matrix
 {
     const Index m = L.rows();
     const Index k = B.cols();
-    Matrix X(m, k);
+    Matrix X = zeros(m, k);
```
tadeu: This helped solve some problems with different `pE` and `Reduction Potential`

tadeu:
Hey @allanleal, I'm investigating the issue of different values in different platforms, on our side, and I've made this small modification that helped with some of these issues. This might help you in https://github.com/reaktoro/reaktoro/pull/88

Reaktoro/Math/LU.cpp
allanleal: Hmmm non-initialized eigen objects could indeed cause these different behaviors. Good catch.

Issue #91, opened 2019-07-09
Residual values when reusing EquilibriumSolver Class


armandoeng:
When you reuse the Equilibrium Solver Class to solve EquilibriumProblem, It looks like it keeps some residual values from last solution, which gives different values.

Below is a script that shows this problem. It has to do a diff on the output results.
If is set the warmstart = False on the EquilibriumSolver options, then that problem doesn't happen.

armandoeng:
```
from reaktoro import *

def create_system():
    database = Database("supcrt07.xml")

    editor = ChemicalEditor(database)
    editor.addAqueousPhaseWithElementsOf(['C', 'Ca', 'Cl', 'Fe', 'H', 'Na', 'O', 'S', 'Ba', 'Sr'])
    editor.addMineralPhase("Anhydrite")
    editor.addMineralPhase("Barite")
    editor.addMineralPhase("Calcite")
    editor.addMineralPhase("Celestite")
    editor.addMineralPhase("Siderite")
    editor.addMineralPhase("Pyrrhotite")

    return ChemicalSystem(editor)


def create_default_problem(system):
    pressure = 13660337.317389656 #Pa
    temperature = 64 #C

    h2o = 394.76384905586252 #kg

    co2 = 25.350676739651345 #mol
    hco3 = 40.7007440302866 #mol
    co3 = 0.026502456032096638 #mol
    h2s = 5.7146433306057664e-24 #mol
    hs = 6.0271635093622933e-24 #mol
    h_p = 0.00026713431941822389 #mol
    oh_m = 0.00070239234495091823 #mol

    so4 = 0.006002566648441959 #kg
    ca = 0.011204791077091658 #kg
    na = 3.1086937041933971 #kg
    cl = 3.3645158450242447 #kg
    fe = 0.00023317970573647533 #kg
    ba = 0.0080034221979226137 #kg
    sr = 0.017607528835429747 #kg

    calcite = 100000000 #kg
    siderite = 0.01795172676219817 #mol

    inputs = {
        'H2O(l)': (h2o, 'kg'),
        'CO2(aq)': (co2, 'mol'),
        'HCO3-': (hco3, 'mol'),
        'CO3--': (co3, 'mol'),
        'H2S(aq)': (h2s, 'mol'),
        'HS-': (hs, 'mol'),
        'H+': (h_p, 'mol'),
        'OH-': (oh_m, 'mol'),
        'SO4--': (so4, 'kg'),
[...]

PR #93, opened 2019-08-02
Fix issue #91

allanleal:
This PR is a fix for issue #91.


Reaktoro/Thermodynamics/Core/Database-test.py
```diff
@@ -109,7 +109,7 @@ def test_locale_problem_with_pugixml():
             check_molar_mass({element.name(): element.molarMass() for element in database.elements()})
 
     assert at_least_one_valid_locale, "Couldn't find any valid locale to test"
-    assert at_least_one_locale_has_comma_for_decimal_separator, "Couldn't find any locale with comma for decimal separator"
+    # assert at_least_one_locale_has_comma_for_decimal_separator, "Couldn't find any locale with comma for decimal separator"
```
allanleal: I had to comment this, because it fails in my machine. There is no locale here with comma as decimal separator.

tests/python_tools.py
```diff
@@ -10,6 +10,8 @@ def convert_reaktoro_state_to_dict(state):
     n = state.speciesAmounts()
     y = state.elementDualPotentials()
     z = state.speciesDualPotentials()
+    b = state.elementAmounts()
+    R = 8.314462618  # universal gas constant in J/(mol*K)
```
tadeu: This seems to be already defined here with a different value: https://github.com/reaktoro/reaktoro/blob/d8bbb227b31e7a68004151ce118651134fbc68db/Reaktoro/Common/Constants.hpp#L23

I think it doesn't need to be in this PR, but maybe all R definitions in the repo (C++ and Python) can be standardized for the new CODATA 2018 value (`8.31446261815324`), which now is exact, _by definition_, and hopefully won't change every two years anymore :)

The difference is minimal (much lower than other sources of uncertainties), but it's one less thing to worry about when comparing numerical differences :)

tests/regression/test_different_results.py
```diff
@@ -80,4 +80,4 @@ def test_different_results(state_regression):
 
     assert result.optimum.succeeded
 
-    state_regression.check(chemical_state, default_tol=dict(atol=1e-50, rtol=1e-16))
+    state_regression.check(chemical_state, default_tol=dict(atol=1e-5, rtol=1e-14))
```
tadeu: Just wondering if `1e-5` for absolute tolerance could be too much... I don't remember the order of magnitude of some values, but, for example, if there are values that are significant at the order of 1e-6, then `expected==1e-6, obtained==9e-6` would pass the test, even with this 9x relative difference.

allanleal:
Thanks a lot for the review @tadeu !

tests/regression/test_different_results.py
allanleal: The change from `1e-50` to `1e-5` was motivated because `1e-5` is used for all other tests (abs tol), and thus I thought `1e-50` was a typo. But yes, you are right - this abstol should be slightly more strict (e.g., 1e-12). I just hope this will not lead to issues across different systems! :)

tests/python_tools.py
allanleal: Excellent - I did not know about this recent standard value. It makes sense indeed to use such value and search for occurrences of `8.314\d+` and replace the hard wired values by `universalGasConstant` with value updated to `8.31446261815324`.

PR #99, opened 2019-08-20
Add python bindings for ThermoData and tests

armandoeng:
- replaced reset function to operator= on class Optional - it was deprecated
- To avoid implementing a pybind to Optional class, it was used lambda functions to properly handle it
- It was created a simple get function to handle the optional state
- The tests were very simple, just to test the pybind implementation


python/PyReaktoro/Math/PyBilinearInterpolator.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+
+    py::class_<BilinearInterpolator>(m, "BilinearInterpolator")
+            .def(py::init<>())
+            .def(py::init<const std::vector<double>&, const std::vector<double>&, const std::vector<double>&>())
+            .def(py::init<const std::vector<double>&, const std::vector<double>&, const std::function<double(double, double)>&>())
+            .def("setCoordinatesX", &BilinearInterpolator::setCoordinatesX)
+            .def("setCoordinatesY", &BilinearInterpolator::setCoordinatesY)
+            .def("setData", &BilinearInterpolator::setData)
+            .def("xCoodinates", &BilinearInterpolator::xCoodinates, py::return_value_policy::reference_internal)
```
tadeu: There is a typo in the method name here and below (`xCoodinates` -> `xCoordinates`). It was already in the code, but it's better to not propagate it (it was being used in very few places, should be easy to find & replace).

Reaktoro/Math/BilinearInterpolator-test.py
```diff
@@ -0,0 +1,44 @@
...
+    data = [4, 5, 6, 7,
+            5, 5, 6, 7,
+            6, 6, 6, 7,
+            7, 7, 7, 7]
+    empty_interpolator.setCoordinatesX(coordinates)
+    empty_interpolator.setCoordinatesY(coordinates)
+    empty_interpolator.setData(data)
+
+    interpolator = BilinearInterpolator(coordinates, coordinates, data)
+    interpolator.data() #dummy call to test any problem with losing data
```
tadeu: What is meant by "losing data"?

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,143 @@
...
+#include <Reaktoro/Thermodynamics/Species/ThermoData.hpp>
+
+namespace Reaktoro {
+
+namespace {
+
+template<typename Property>
+py::object get_optional(Optional<Property> const& property) {
+    if (!property.empty()) {
+        return py::cast<Property>(property.get());
```
tadeu: Would `py::cast(property.get())` work here?

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+}
+
+void exportThermoData(py::module& m)
+{
+
+    py::class_<SpeciesThermoData>(m, "SpeciesThermoData")
+            .def(py::init<>())
+            .def_property("properties",
```
tadeu: 1) Consider not chaining things like this, but using this idiom, it's usually better for code maintenance, debugging and error tracking:

```cpp
    {
        using T = SpeciesThermoData;
        auto c = py::class_<T>(m, "SpeciesThermoData");
        c.def(py::init<>());
        c.def_property(...);
        ...
    }
```

2) Indentation seems a little bit inconsistent (sometimes 4, sometimes 8, etc.)

Reaktoro/Math/BilinearInterpolator-test.py
armandoeng: As talked, improved the comment to explain that in some case, the vector could be moved and deleted, and it need to call it twice to get it.

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,143 @@
...
+    py::class_<SpeciesThermoParamsPhreeqc>(m, "SpeciesThermoParamsPhreeqc")
+            .def(py::init<>())
+            .def_readwrite("reaction", &SpeciesThermoParamsPhreeqc::reaction);
+
+    py::class_<SpeciesThermoParamsPhreeqc::ReactionParams>(m, "ReactionParams")
+            .def(py::init<>())
+            .def_readwrite("equation", &SpeciesThermoParamsPhreeqc::ReactionParams::equation)
+            .def_readwrite("log_k", &SpeciesThermoParamsPhreeqc::ReactionParams::log_k)
+            .def_readwrite("delta_h", &SpeciesThermoParamsPhreeqc::ReactionParams::delta_h)
+            .def_readwrite("analytic", &SpeciesThermoParamsPhreeqc::ReactionParams::analytic, py::return_value_policy::reference_internal);
```
tadeu: I think it would be better to have tests for all these bindings, especially ones like this that are not "straightforward" (a `vector` bound using `reference_internal`).

Current tests seem way too simple, testing only creation with default constructors, and setting _some_ things.

Reaktoro/Math/BilinearInterpolator-test.py
```diff
@@ -0,0 +1,48 @@
...
+    interpolator = BilinearInterpolator(coordinates, coordinates, data)
+    assert not interpolator.empty()
+
+    #it is needed to call data() twice to verify if the vector is not gonna be moved to interpolator_data and then destroyed
+    interpolator_data = interpolator.data()
+    assert all(i == j for i, j in zip(interpolator_data, data))
+    assert all(i == j for i, j in zip(empty_interpolator.data(), data))
+    assert interpolator.xCoordinates()
+    assert interpolator.yCoordinates()
+    assert interpolator(1.5, 1.5) == empty_interpolator(1.5, 1.5)
```
allanleal: We could also have:

~~~
assert interpolator(0, 0) == 4
assert interpolator(1, 1) == 5
assert interpolator(2, 2) == 6
assert interpolator(3, 3) == 7
assert interpolator(0.5, 0.0) == 4.5
assert interpolator(2.5, 2.5) == 6.5
~~~

Reaktoro/Math/BilinearInterpolator.cpp
```diff
@@ -99,12 +99,12 @@ auto BilinearInterpolator::setData(const std::vector<double>& data) -> void
     m_data = data;
 }
 
-auto BilinearInterpolator::xCoodinates() const -> const std::vector<double>&
+auto BilinearInterpolator::xCoordinates() const -> const std::vector<double>&
```
allanleal: :+1: :)

python/PyReaktoro/Math/PyBilinearInterpolator.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <Reaktoro/Math/BilinearInterpolator.hpp>
+
+namespace Reaktoro {
+
+void exportBilinearInterpolator(py::module& m)
+{
+
+    py::class_<BilinearInterpolator>(m, "BilinearInterpolator")
+            .def(py::init<>())
```
allanleal: Please indent using 4 spaces (possibly using tab here)

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
allanleal: Shall we start using CXX17 and then remove this `Optional` type, which is a wrapper to `boost::optional`, and use directly `std::optional`?

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
allanleal: I agree with @tadeu that *cascading calls* can sometimes be problematic, but it has rarely been the case using Boost.Python in the past and now pybind11; both tools also document usage using this style, and the entire code for python bindings in the project already follows this convention, so to be consistent, I would propose we continue with this style for now.

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,143 @@
...
+                          [](GaseousSpeciesThermoData const& self) { return get_optional(self.hkf); },
+                          [](GaseousSpeciesThermoData& self, GaseousSpeciesThermoParamsHKF property) { self.hkf.set(property); });
+
+    py::class_<MineralSpeciesThermoData, SpeciesThermoData>(m, "MineralSpeciesThermoData")
+            .def(py::init<>())
+            .def_property("hkf",
+                          [](MineralSpeciesThermoData const& self) { return get_optional(self.hkf); },
+                          [](MineralSpeciesThermoData& self, MineralSpeciesThermoParamsHKF property) { self.hkf.set(property); });
+
+}
```
allanleal: These bindings are more complex than should because we have this `Optional` wrapper. Once replaced by `std::optional`, we should only have `.def_readwrite(...)` statementes. Also, we do not need `.def(py::init<>())` for POD `struct`; I think those types no even implement constructors.

allanleal: Attention to the indentation length too! :)

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
allanleal: Good observation here, @tadeu . Why do we need `py::return_value_policy::reference_internal` for a read-write member field? AFAIK, we would have it for the return of a member method when this returns a reference to internal data.

tadeu: Indeed, it may not be necessary, and it may be the case that at every access the whole vector is being copied (converted to/from Python list). @armandoeng, perhaps in this case it's better to use an Opaque Type, what do you think? Take a look here: https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#making-opaque-types

python/PyReaktoro/Math/PyBilinearInterpolator.cpp
armandoeng: Oops.

armandoeng:
Did a squash, mainly because one of my commits changed the newline from some files.

Reaktoro/Thermodynamics/Core/ChemicalEditor.cpp
```diff
@@ -1,601 +1,602 @@
-// Reaktoro is a unified framework for modeling chemically reactive systems.
```
allanleal: Something happened here that the diff of `ChemicalEditor.cpp` is not showing the individual changes (the entire file is changed, but look very similar).

Reaktoro/Thermodynamics/Core/Database.cpp
```diff
@@ -1,831 +1,830 @@
-// Reaktoro is a unified framework for modeling chemically reactive systems.
```
allanleal: The same here, and possibly also for a few other files below where the diff is long and not shown directly by GitHub.

Reaktoro/Thermodynamics/Species/ThermoData.hpp
```diff
@@ -62,37 +62,13 @@ struct SpeciesThermoInterpolatedProperties
 /// The thermodynamic properties are represented by BilinearInterpolator instances, where
 /// the temperature points (in units of K) are along the x-coordinates, and the pressure points
 /// (in units of Pa) are along the y-coordinates.
-struct ReactionThermoInterpolatedProperties
+struct ReactionThermoInterpolatedProperties : public SpeciesThermoInterpolatedProperties
```
allanleal: For the same reason we don't (at least most of C++ literature) do:

~~~c++
struct Foo 
{
public:
    int a;
    char b;
};
~~~

but instead just:

~~~c++
struct Foo 
{
    int a;
    char b;
};
~~~

because `struct` has public access by default, we should also do:

~~~c++
struct Foo : Bar // not struct Foo : public Bar 
{
...
};
~~~

since `struct` also by default inherit with `public` access to parent's public members.

python/PyReaktoro/PyReaktoro.hpp
```diff
@@ -108,6 +109,8 @@ void exportAqueousSpecies(py::module& m);
 void exportGaseousSpecies(py::module& m);
 void exportMineralSpecies(py::module& m);
 void exportWater(py::module& m);
+void exportThermoData(py::module& m);
+void exportThermoDataProperties(py::module& m);
```
allanleal: Better to merge this PR after the other one where you fix the bug when inert elements exist.

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,118 @@
...
+        .def_readwrite("gibbs_energy", &SpeciesThermoInterpolatedProperties::gibbs_energy)
+        .def_readwrite("helmholtz_energy", &SpeciesThermoInterpolatedProperties::helmholtz_energy)
+        .def_readwrite("internal_energy", &SpeciesThermoInterpolatedProperties::internal_energy)
+        .def_readwrite("enthalpy", &SpeciesThermoInterpolatedProperties::enthalpy)
+        .def_readwrite("entropy", &SpeciesThermoInterpolatedProperties::entropy)
+        .def_readwrite("volume", &SpeciesThermoInterpolatedProperties::volume)
+        .def_readwrite("heat_capacity_cp", &SpeciesThermoInterpolatedProperties::heat_capacity_cp)
+        .def_readwrite("heat_capacity_cv", &SpeciesThermoInterpolatedProperties::heat_capacity_cv);
+
+    py::class_<ReactionThermoInterpolatedProperties, SpeciesThermoInterpolatedProperties>(m,
```
allanleal: Better to keep everything in the same line, or just indent 4 spaces (instead of aligning the arguments with `(`).

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,118 @@
...
+        .def_readwrite("Vr", &MineralSpeciesThermoParamsHKF::Vr)
+        .def_readwrite("nptrans", &MineralSpeciesThermoParamsHKF::nptrans)
+        .def_readwrite("a", &MineralSpeciesThermoParamsHKF::a)
+        .def_readwrite("b", &MineralSpeciesThermoParamsHKF::b)
+        .def_readwrite("c", &MineralSpeciesThermoParamsHKF::c)
+        .def_readwrite("Ttr", &MineralSpeciesThermoParamsHKF::Ttr)
+        .def_readwrite("Htr", &MineralSpeciesThermoParamsHKF::Htr)
+        .def_readwrite("Vtr", &MineralSpeciesThermoParamsHKF::Vtr)
+        .def_readwrite("dPdTtr", &MineralSpeciesThermoParamsHKF::dPdTtr)
+        .def_readwrite("Tmax", &MineralSpeciesThermoParamsHKF::Tmax);
```
allanleal: There is some convenience (and symmetry) in placing `;` on a new line. In case later we need to add one or more new bindings we don't need to be concerned about moving `;` to the last binding call.

allanleal:
Looking good @armandoeng ! I'm just checking as you go, so that we agree/discuss on the changes earlier rather than at the very. ;)

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,118 @@
...
+        .def(py::init<>())
+        .def_readwrite("reaction", &SpeciesThermoParamsPhreeqc::reaction);
+
+    py::class_<SpeciesThermoParamsPhreeqc::ReactionParams>(m, "ReactionParams")
+        .def(py::init<>())
+        .def_readwrite("equation", &SpeciesThermoParamsPhreeqc::ReactionParams::equation)
+        .def_readwrite("log_k", &SpeciesThermoParamsPhreeqc::ReactionParams::log_k)
+        .def_readwrite("delta_h", &SpeciesThermoParamsPhreeqc::ReactionParams::delta_h)
+        .def_readwrite("analytic", &SpeciesThermoParamsPhreeqc::ReactionParams::analytic,
+                       py::return_value_policy::reference_internal);
```
allanleal: Can we remove `py::return_value_policy::reference_internal` here?

python/PyReaktoro/PyReaktoro.cpp
```diff
@@ -115,6 +120,8 @@ using namespace Reaktoro;
 
 PYBIND11_MODULE(PyReaktoro, m)
 {
+    py::bind_vector<std::vector<double>>(m, "VectorDouble", "VectorDouble Descriptor");
```
allanleal: Shouldn't we rely only on `numpy` and Python native types (e.g., list, dict, tuple, etc.). I mean, ideally, we should not have to deal with `std::vector<double>` from Python side. Can we avoid this?

python/PyReaktoro/Thermodynamics/Species/PyThermoData .cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    py::class_<SpeciesThermoData>(m, "SpeciesThermoData")
+        .def(py::init<>())
+        .def_readwrite("properties", &SpeciesThermoData::properties)
+        .def_readwrite("reaction", &SpeciesThermoData::reaction)
+        .def_readwrite("phreeqc", &SpeciesThermoData::phreeqc)
+        ;
+
+    py::class_<AqueousSpeciesThermoData, SpeciesThermoData>(m, "AqueousSpeciesThermoData")
+        .def(py::init<>())
```
allanleal: We don't need `.def(py::init<>())` for `struct` types that do not implement custom default constructors. I'll quickly edit this file.

tests/test_reaktoro_misc.py
```diff
@@ -34,3 +34,20 @@ def test_dict_on_chemical_state(create_minimum_system):
 def test_dict_on_chemical_quantity(create_minimum_system):
     quantity = ChemicalQuantity(create_minimum_system)
     check_dict(quantity)
+
+
+def test_VectorDouble():
```
allanleal: Let's try not to expose std::vector<double> to Python, since pybind11 has amazing conversion capabilities/interop between C++ STL and Python native types.

allanleal:
Hi Armando. I made a few more revisions, and some quick edits for you. Please have a look at the new comments.

Reaktoro/Math/BilinearInterpolator.cpp
```diff
@@ -128,20 +128,19 @@ auto BilinearInterpolator::operator()(double x, double y) const -> double
     const double xB = m_xcoordinates.back();
     const double yA = m_ycoordinates.front();
     const double yB = m_ycoordinates.back();
+    if(x > xB || x < xA || y > yB || y < yA)
+        interpolationOutOfBoundsError(x, xA, xB, y, yA, yB);
```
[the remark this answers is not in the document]
allanleal: Excellent.

Issue #101, opened 2019-08-22
Not setting element amounts vector when setting an new Partition on EquilibriumSolver


armandoeng:
When do you have an inert element on partition, the solver complains:

```
E       RuntimeError: 
E       ***************************************************************************************************************************************************************
E       *** Error: Cannot proceed with method EquilibriumSolver::solve.
E       *** Reason: The dimension of the given vector of molar amounts of the elements does not match the number of elements in the equilibrium partition.
E       *** Location:  This error was encountered in Reaktoro/Equilibrium/EquilibriumSolver.cpp:450.
E       ***************************************************************************************************************************************************************
```

Test to get that error on python, using the tests on EquilibriumSolver-test:
```
    editor = ChemicalEditor()
    editor.addAqueousPhase("H2O(l) H+ OH- HCO3- CO2(aq) CO3--".split())
    editor.addGaseousPhase("H2O(g) CO2(g) Ar(g)".split())
    editor.addMineralPhase("Graphite")
    chemical_system_adding_argon = ChemicalSystem(editor)
    partition = Partition(chemical_system_adding_argon)
    partition.setInertPhases(['Gaseous'])
    problem = _create_equilibrium_problem(partition)
    state = _create_chemical_state(chemical_system_adding_argon)

    # Compute equilibrium state; the amounts of CO2(g) and H2O(g) should remain the same
    solver = EquilibriumSolver(chemical_system_adding_argon)
    solver.setPartition(problem.partition())
    solver.solve(state, problem)
```

I don't know if EquilibriumProblem should return a elementAmounts with the corrected size, or if that is something the EquilibriumSolver should handle.

armandoeng:
My fix for this issue: https://github.com/reaktoro/reaktoro/pull/102

PR #102, opened 2019-08-22
EquilibriumSolver::solve fails with an EquilibriumProblem that has inert elements

armandoeng:
Issue #101


Reaktoro/Equilibrium/EquilibriumProblem.cpp
```diff
@@ -270,9 +270,9 @@ auto EquilibriumProblem::pressure() const -> double
     return pimpl->P;
 }
 
-auto EquilibriumProblem::elementAmounts() const -> VectorConstRef
+auto EquilibriumProblem::elementAmounts() const -> Vector
```
tadeu: It looks like, in this way, `EquilibriumProblem::elementAmounts()` is not consistent with `EquilibriumProblem::setElementAmounts(VectorConstRef b)`.

allanleal:
Currently, `EquilibriumProblem::elementAmounts` return a vector with element amounts of all elements, agnostic to partition. However, `EquilibriumSolver::solve` expects for argument `be` a vector of element amounts of the elements in the equilibrium partition (thus, it can be shorter than that returned by   `EquilibriumProblem::elementAmounts`.

This will introduce a breaking change, but I think we should have `EquilibriumSolver::solve` expecting a `b` argument (full length vector with amounts of all elements) and internally, it figures out the removal of those that matter for the equilibrium calculation for a selection of species that are under equilibrium (i.e., the *equilibrium partition*).

armandoeng:
> This will introduce a breaking change, but I think we should have `EquilibriumSolver::solve` expecting a `b` argument (full length vector with amounts of all elements) and internally, it figures out the removal of those that matter for the equilibrium calculation for a selection of species that are under equilibrium (i.e., the _equilibrium partition_).

Totally agree. To reinforce that argument, not only the changes on EquilibriumProblem fells strange, but as EquilibriumSolver accepts as well a ChemicalState, that last would require to have the same intervention (and it doesn't has a partition).

python/PyReaktoro/PyReaktoro.cpp
```diff
@@ -16,6 +16,101 @@
 // along with this library. If not, see <http://www.gnu.org/licenses/>.
 
 #include <PyReaktoro/PyReaktoro.hpp>
+
```
allanleal: These changes here and below (on `PyReaktoro.hpp` and `PyReaktoro.cpp`) have nothing to do with the issue. They are needed only to avoid unnecessary build of all python binding files (the `PyClassName.cpp` files) when any change is made in `PyReaktoro.cpp` (even a white space would trigger the entire compilation of all such files).

armandoeng:
I did a squash to organize the commits. 
@allanleal, as it has to intervene on approximate as well, I did some refactoring. See if its good to you. :)

Reaktoro/Equilibrium/EquilibriumSolver.cpp
```diff
@@ -625,12 +642,13 @@ auto EquilibriumSolver::solve(ChemicalState& state, double T, double P, const do
 
 auto EquilibriumSolver::solve(ChemicalState& state) -> EquilibriumResult
 {
-    return solve(state, state.temperature(), state.pressure(), state.elementAmounts());
+    return pimpl->solve_with_all_elements(state, state.temperature(), state.pressure(),  state.elementAmounts());
```
allanleal:
```suggestion
    return pimpl->solve_with_all_elements(state, state.temperature(), state.pressure(), state.elementAmounts());
```

Reaktoro/Equilibrium/EquilibriumSolver.cpp
```diff
@@ -625,12 +642,13 @@ auto EquilibriumSolver::solve(ChemicalState& state, double T, double P, const do
...
 }
 
 auto EquilibriumSolver::solve(ChemicalState& state, const EquilibriumProblem& problem) -> EquilibriumResult
 {
-    return solve(state, problem.temperature(), problem.pressure(), problem.elementAmounts());
+    setPartition(problem.partition());
+    return pimpl->solve_with_all_elements(state, problem.temperature(), problem.pressure(),  problem.elementAmounts());
```
allanleal:
```suggestion
    return pimpl->solve_with_all_elements(state, problem.temperature(), problem.pressure(), problem.elementAmounts());
```

Reaktoro/Equilibrium/EquilibriumSolver.cpp
```diff
@@ -439,6 +446,13 @@ struct EquilibriumSolver::Impl
         return zero || !options.warmstart;
     }
 
+    /// Solve the equilibrium problem, passing all elements that has on chemical system
+    auto solve_with_all_elements(ChemicalState& state, double T, double P, VectorConstRef b) -> EquilibriumResult
```
allanleal: `solve_with_all_element_amounts` would be more preferable.

Reaktoro/Equilibrium/EquilibriumSolver.cpp
```diff
@@ -595,7 +609,9 @@ auto EquilibriumSolver::setOptions(const EquilibriumOptions& options) -> void
 
 auto EquilibriumSolver::setPartition(const Partition& partition) -> void
 {
-    pimpl->setPartition(partition);
+    if(partition != pimpl->partition){
```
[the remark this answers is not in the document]
armandoeng: Agree. I will delete the implemented operator== on partition, as it will have no use now.

PR #109, opened 2019-09-02
Fix missing propagation of CXX17 requirement to dependent codes

allanleal:
This PR aims to fix a current issue that dependent code (using `find_package(Reaktoro)` and linking using `target_link_libraries(mylib Reaktoro::Reaktoro)`) does not get automatically the CXX17 standard requirement. By transferring this compilation requirement to the library target `Reaktoro`, this fixes the issue.


Reaktoro/CMakeLists.txt
```diff
@@ -60,7 +63,7 @@ endif()
 # Install Reaktoro header files
 install(DIRECTORY ${PROJECT_SOURCE_DIR}/Reaktoro
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR} COMPONENT headers
-    FILES_MATCHING PATTERN "*.hpp")
+    FILES_MATCHING PATTERN "*.hpp" PATTERN "*.hxx")
```
allanleal: `.hxx` files were not being copied to the install dir.

Reaktoro/Core/ChemicalProperties.cpp
```diff
@@ -62,7 +62,16 @@ auto ChemicalProperties::update(VectorConstRef n_) -> void
...
+        if(size == 1) {
+            xp = 1.0;
+        }
+        else {
+            const auto snpc = sum(npc);
+            if(snpc != 0.0)
+                xp = npc/snpc;
+            else
+                xp = 0.0;
+        }
```
allanleal: These alterations resolve division by zero when computing mole fractions for phases that have zero amount for all species.

Reaktoro/CMakeLists.txt
```diff
@@ -40,6 +40,9 @@ target_link_libraries(Reaktoro
     PRIVATE ${THIRDPARTY_LIBS}
     PUBLIC Boost::boost)
 
+# Set compilation features to be propagated to dependent codes.
+target_compile_features(Reaktoro PUBLIC cxx_std_17)
```
tadeu: Nice, today I learned about this other way of configuring C++17 feaures, being better because it propagates :)

I was reading this: https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html#cmake-38-meta-compiler-features

They also recommend `set_target_properties(myTarget PROPERTIES CXX_EXTENSIONS OFF)` so that the code is more portable, what do you think?

Also, as always there is still _another_ way to configure this 🙄:

https://cmake.org/cmake/help/v3.15/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

CMakeLists.txt
```diff
@@ -59,12 +59,6 @@ if(NOT CMAKE_BUILD_TYPE)
     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release MinSizeRel RelWithDebInfo)
 endif()
 
-# Set libraries to be compiled with position independent code mode (i.e., fPIC option in GNU compilers)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
tadeu: Could removing this line cause problems when compiling static libraries into shared libraries? (i.e., first Reaktoro is built as static, then it is linked into another shared library)

allanleal: I think only shared libs need `-fPIC`. From [cmake docs](https://cmake.org/cmake/help/v3.0/command/add_library.html):

`For SHARED and MODULE libraries the POSITION_INDEPENDENT_CODE target property is set to ON automatically.`

I think the way it was we could have issues if static libs are built.

Reaktoro/CMakeLists.txt
allanleal: I think what you are pointing out here could be useful indeed. pybind11 seemed to overwriting c++17 with c++14. Both osx and linux builds in travis were passing. But the win build on appveyor was failing when the py bindings were being built and `std:: optional` undeclared errors popping up.

CMakeLists.txt
tadeu: From here:

https://stackoverflow.com/questions/3961446/why-does-gcc-not-implicitly-supply-the-fpic-flag-when-compiling-static-librarie

> It depends on what use you will have for your static library. If you only want to link it into programs, it doesn't need PIC code (libtool calls that a convenience library, because you could pretty much do without it, it simply helps get your compilation process to a reasonable size, for example). *Otherwise, if you intend to link shared libraries against it, you need PIC code in your static library.*

But I think it's also ok if you want to leave the default, if someone stumbles upon this scenario, CMake can be configured from the command line.

allanleal: You're right! So, I have added this back in a more target-based fashion rather than global option:

~~~cmake
set_target_properties(Reaktoro PROPERTIES
    POSITION_INDEPENDENT_CODE ON
    CXX_EXTENSIONS OFF)
~~~

PR #116, opened 2019-09-11
fix pybind bug where move dynamic content from optional attributes

armandoeng:
Bug related to the issue: https://github.com/pybind/pybind11/issues/1919

### DIFF
--- a/python/PyReaktoro/Thermodynamics/Species/PyThermoData.cpp
+++ b/python/PyReaktoro/Thermodynamics/Species/PyThermoData.cpp
@@ -21,32 +21,52 @@
 
 namespace Reaktoro {
 
-void exportThermoData(py::module& m) 
+//TODO: Fix for pybind bug, where it moves optional attributes, losing its dynamic content.
+//https://github.com/pybind/pybind11/issues/1919
+template <typename _Property>
+_Property* get_optional(std::optional<_Property>& prop)
+{
+    if (prop.has_value())
+        return &prop.value();
+        
+    return nullptr;
+}
+
+void exportThermoData(py::module& m)
 {
     py::class_<SpeciesThermoData>(m, "SpeciesThermoData")
         .def(py::init<>())
-        .def_readwrite("properties", &SpeciesThermoData::properties)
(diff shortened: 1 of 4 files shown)


python/PyReaktoro/Thermodynamics/Species/PyThermoData.cpp
```diff
@@ -21,32 +21,52 @@
...
     m.attr("LiquidSpeciesThermoData") = m.attr("FluidSpeciesThermoData");
     m.attr("GaseousSpeciesThermoData") = m.attr("FluidSpeciesThermoData");
-
-    py::class_<MineralSpeciesThermoData, SpeciesThermoData>(m, "MineralSpeciesThermoData")
+    
+    {
+        py::class_<MineralSpeciesThermoData, SpeciesThermoData> (m, "MineralSpeciesThermoData")
         .def(py::init<>())
-        .def_readwrite("hkf", &MineralSpeciesThermoData::hkf)
+        .def_property("hkf", [&] (MineralSpeciesThermoData& self) { return get_optional(self.hkf);},
```
tadeu: is `[&]` necessary here (as opposed to plain no-capture `[]`)?

tadeu: Ah first I thought that `keep_alive` would be necessary here, according to the [docs](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies), the default return value policy for pointers is:

> `return_value_policy::take_ownership`: Reference an existing object (i.e. do not create a new copy) and take ownership. **Python will call the destructor and delete operator when the object’s reference count reaches zero**. Undefined behavior ensues when the C++ side does the same, or when the data was not dynamically allocated.

which seems dangerous. But then, hidden in the `return_value_policy::reference_internal` description it says:

> This is the default policy for property getters created via def_property, def_readwrite, etc.

So I guess it's ok to use the default 👍

armandoeng: Lefts from my despair.

PR #117, opened 2019-09-17
Add partial molar volumes on Chemical Properties

armandoeng:


Reaktoro/Core/ChemicalProperties.hpp
```diff
@@ -88,6 +88,9 @@ class ChemicalProperties
     /// Return the ln activities of the species.
     auto lnActivities() const -> ChemicalVectorConstRef;
 
+    /// Return the partial molar volume of the species in that phase (in units of m3/mol).
+    auto partialMolarVolumes() const->ChemicalVectorConstRef;
```
allanleal:
```suggestion
    auto partialMolarVolumes() const -> ChemicalVectorConstRef;
```

PR #120, opened 2019-10-08
Fix optional value compilation error clang9 macos

allanleal:
This PR fixes the compilation issues in conda-forge builds for macOS. This has to do with method `std::optional::value` not being supported on clang9 for macOS. All such uses were replaced by `std::optional::operator*`.

See [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=78862) for more details about the error in the compilation log.

The solution was taken from this [stackoverflow thread](https://stackoverflow.com/questions/44217316/how-do-i-use-stdoptional-in-c?answertab=active#tab-top).


allanleal:
I'll merge this now (all checks passed) as it is just a minor fix (although several files affected).

PR #124, opened 2019-10-16
WIP Fix some cmake issues when compiling without python and using intel compiler

allanleal:
This PR fixes the following issues:

* `-DREAKTORO_BUILD_PYTHON=FALSE` still requires `pybind11` to be found.
* MSVC compile options should also be used when compiling with Intel compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # Set the list of compiler flags for MSVC compiler
-if(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC OR ${CMAKE_CXX_COMPILER_ID} STREQUAL Intel)
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
     add_compile_options(
         /D_SCL_SECURE_NO_WARNINGS
         /D_CRT_SECURE_NO_WARNINGS=1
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -102,38 +100,22 @@ endif()
 
 # Build the demonstration applications
 if(REAKTORO_BUILD_DEMOS)
-    add_subdirectory(demos)
-else()
     add_subdirectory(demos EXCLUDE_FROM_ALL)
```
tadeu: Should `EXCLUDE_FROM_ALL` be removed from here? (and also from the changes below)

allanleal: Maybe yes. My idea at the moment is that by default `make`, or `ninja`, builds the shared library and the python module. To build the other components, we would do `make demos`, `make docs`, etc. To disable these components all together, we could instead have cmake variables named `REAKTORO_DISABLE_PYTHON`, `REAKTORO_DISABLE_DOCS`. This is still not clear - but I just wanted to reply here now rather than later. :)

CMakeLists.txt
```diff
@@ -72,6 +72,12 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
     )
 endif()
 
+# Set the list of compiler flags for Intel compiler
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL Intel)
+    add_compile_options(
```
tadeu: Is this supposed to be empty?

allanleal:
Sorry, Tadeu, I forgot to add a WIP tag in this PR. It's empty because I
still don't know exactly the options to use for intel compiler.

Issue #16, opened 2018-10-11
Use pre-commit and black for formatting


nicoddemus:
Want to get opinions if we should start to use [pre-commit](https://pre-commit.com/) for general checkins (like docs, formatting, etc) and [black](https://github.com/ambv/black) for automatic code formatting of Python code.

I've been using both in all my open source projects and have been happy with it. More and more Python projects are also adopting black as their auto formatter.

allanleal:
Hi @nicoddemus . Thanks for the suggestion. The black project seems interesting, and its adoption in Reaktoro will ensure everyone contributing python code does not need to be super concerned about their formatting as this gets corrected automatically. So, we have consistent format across all python codes.

For C++, I know something similar can be achieved with clang. Have you used clang before for formatting C++ codes? We should consider this as well.

I still need a bit more time to read about the pre-commit. I'll get back soon.

nicoddemus:
> For C++, I know something similar can be achieved with clang. Have you used clang before for formatting C++ codes?

AFAIK we never used a formatting tool for C++, but @tadeu will know more details. 👍

allanleal:
Hi @nicoddemus. The pre-commit (and potentially other git hooks) seems like a nice idea. Could you please elaborate a bit more on what we would achieve with pre-commit (a simple/short example)?

nicoddemus:
`pre-commit` allows you to run scripts or tools during each commit to verify syntax, formatting, validation, anything you might want really. This might prevent errors from being introduced in a PR and helps keeping the codebase uniform.

`pre-commit` itself contains a lot of [out-of-the-box hooks](https://github.com/pre-commit/pre-commit-hooks), I think the ones that we could use in Reaktoro are:

* `black` - For Python code formatting
* `blacken-docs` - Ensures code snippets in the docs are also formatted with black
* `end-of-file-fixer` - Makes sure files end in a newline and only a newline.
* `mixed-line-ending` - Replaces or checks mixed line ending (unfortunately this happens more often than it should)
* `trailing-whitespace` - Trims trailing whitespace.

We can also investigate linters for Python and C++ as well, we would need to investigate those.

allanleal:
I see, sounds good to me. Is this something easy to set up for our other repos as well (not necessarily needed now)?

nicoddemus:
Yep, usually is just placing a `.pre-commit-config.yaml` file in the repository and update the developer documentation. 😁

allanleal:
Cool! :)

[...]

PR #127, opened 2019-11-03
Add pre commit and formatters

volpatto:
Hi, @allanleal and `reaktoro`'s contributors!

This PR aims to include pre-commit, providing several git hooks for a proper standardization while developing `reaktoro`. The PR is based on #16.

Following the issue, the below modifications are submitted:

* `pre-commit` is included in `environment.devenv.yaml`. In order to use it, all developers must have `pre-commit` installed, then run in the console:
```console
$ pre-commit install
```

After that, git hooks are settled (and will be performed before every commit), but the developer can run the following command, just to guarantee that everything is according to the formatters:
```console
$ pre-commit run --all-files
```

If the developer wants to check just one of the new hooks, the developer can execute:
```console
$ pre-commit run <id> --all-files
```

The `id` means the hook id, one can find all the available ids inside `.pre-commit-config.yaml`. For example, if you want to run just `black` in all files, you can do `$ pre-commit run black --all-files`.

* The main modifications are the files `.pre-commit-config.yaml`, `pyproject.toml` (this one provides additional configurations to `black` formatter) and `environment.devenv.yaml` (which includes `pre-commit` and `black`). All the modifications which are not in these files were done by the formatters. Please note that the diff is huge!

* The following hooks are added:
    - trailing-whitespace
    - end-of-file-fixer
    - black
    - blacken-docs

Information about the hooks can be found [here](https://pre-commit.com/hooks.html). I need help to configure `clang-format`. I never did it before and I have been struggling with it.


volpatto:
Just to emphasize, I need help with `clang-format` inside `pre-commit`. I never did it before and I'm struggling with it. Could you please help me, @tadeu?

volpatto:
> Just to emphasize, I need help with `clang-format` inside `pre-commit`. I never did it before and I'm struggling with it. Could you please help me, @tadeu?

It's noteworthy that maybe not all the code should be checked and formatted. For example, code inside `Reaktoro/deps` or in `thirdparty`. Of course, no problem at all when using pre-commit, which tackles only "diff" code. But in the "after install" pre-commit scenario, when the check will be applied over all source code, `clang-format` can be a problem.

I was checking that there is a ".gitignore"-like for `clang-format`, which is `clang-format-ignore` file, like [this case](https://github.com/mozilla/gecko-dev/blob/master/.clang-format-ignore) or [this one](https://dxr.mozilla.org/mozilla-central/source/.clang-format-ignore).

volpatto:
In order to guarantee that every change is according to the standards of the formatters, I put in both Travis and AppVeyor commands that execute all the `pre-commit` checks. If a modification is submitted as a PR, the build will fail in the case that the format is not appropriated. Please check this commit [here](https://github.com/reaktoro/reaktoro/pull/127/commits/22756dd765dfec614cb62034de8f9ead1b5e308d).

allanleal:
Hi @volpatto - this is pretty cool contribution. Thanks a lot!

1. I do agree that we should not apply any formatting to the `Reaktoro/deps` and `thirdparty` directories. 

2. I sent a clang-format config to @tadeu a while ago. I'm attaching it here (together with `AqueousMixture.cpp.txt` file, that you need to remove the `.txt`, to have a look at the clang formatted file sample with syntax highlighting).

[clang-format-reaktoro.txt](https://github.com/reaktoro/reaktoro/files/3802794/clang-format-reaktoro.txt)

[AqueousMixture.cpp.txt](https://github.com/reaktoro/reaktoro/files/3802798/AqueousMixture.cpp.txt)

We can further discuss such configs (add/remove/change options). 

As to line lengths, I think the 79-chars rule of PEP-8 does more damage than help in many cases. Sometimes we get long lines because we are using more descriptive function/variable names, and breaking a single line into multiple ones can compromise readability. I would prefer to have longer line lengths (say, 120 chars, or not even force wrapping, but rely instead on soft wrapping functionality of the text editor).

We should also preserve some pre-existing formatting such as in the following case:

~~~diff
 for iphase, phase in enumerate(system.phases()):
-        assert system.indicesSpeciesInPhases([iphase]) == \
-            system.indicesSpecies(names(phase.species()))
+        assert system.indicesSpeciesInPhases([iphase]) == system.indicesSpecies(
+            names(phase.species())
        )
~~~

I think the original version reads better than the automatically formatted one. We can discuss such small things in gitter though.

volpatto:
> Hi @volpatto - this is pretty cool contribution. Thanks a lot!

Thanks! For me, it's a pleasure to provide some contribution to such an amazing project like `reaktoro`.

> 
> 1. I do agree that we should not apply any formatting to the `Reaktoro/deps` and `thirdparty` directories.
> 2. I sent a clang-format config to @tadeu a while ago. I'm attaching it here (together with `AqueousMixture.cpp.txt` file, that you need to remove the `.txt`, to have a look at the clang formatted file sample with syntax highlighting).
> 
> [clang-format-reaktoro.txt](https://github.com/reaktoro/reaktoro/files/3802794/clang-format-reaktoro.txt)
> 
> [AqueousMixture.cpp.txt](https://github.com/reaktoro/reaktoro/files/3802798/AqueousMixture.cpp.txt)
> 
> We can further discuss such configs (add/remove/change options).

Ah, great! I'll have a look locally, just to see how it works.

> As to line lengths, I think the 79-chars rule of PEP-8 does more damage than help in many cases. Sometimes we get long lines because we are using more descriptive function/variable names, and breaking a single line into multiple ones can compromise readability. I would prefer to have longer line lengths (say, 120 chars, or not even force wrapping, but rely instead on soft wrapping functionality of the text editor).

Sure, I do agree with that. Inside `pyproject.toml`, I set the line length to 100, which I understand as not too long nor short. I think we can use 120 chars per line, no problems.

> We should also preserve some pre-existing formatting such as in the following case:
> 
> ```diff
>  for iphase, phase in enumerate(system.phases()):
> -        assert system.indicesSpeciesInPhases([iphase]) == \
> -            system.indicesSpecies(names(phase.species()))
> +        assert system.indicesSpeciesInPhases([iphase]) == system.indicesSpecies(
> +            names(phase.species())
>         )
> ```
> 
> I think the original version reads better than the automatically formatted one. We can discuss such small things in gitter though.

Sure, I think we can set it to some target lines through `clang-format` comments. For instance, please take a look below:

```c++
// clang-format off

(some code here)

// clang-format on
```

This way, `clang-format` is not applied. However, this is easy to do in diffs from now on, but hard to circumvent now, in the first formatters run. Could we proceed despite this fact?

allanleal:
Thanks Diego!

100 chars per line is OK. 

> This way, clang-format is not applied. However, this is easy to do in diffs from now on, but hard to circumvent now, in the first formatters run. Could we proceed despite this fact?

Sure, this approach makes sense!

volpatto:
> Thanks Diego!
> 
> 100 chars per line is OK.
> 
> > This way, clang-format is not applied. However, this is easy to do in diffs from now on, but hard to circumvent now, in the first formatters run. Could we proceed despite this fact?
> 
> Sure, this approach makes sense!

Great, then!

Okay, I managed to add `clang-format` as pre-commit. But I had to remove some config options inside `.clang-format` file. The reason is that locally (Ubuntu 18.04), the supported version available from Ubuntu's repos doesn't have some of the options provided in the file @allanleal attached above in a previous comment.

An important point is that `clang-format` will not apply on `Reaktoro/deps` and `thirdparty`. Please take a look in the file `.clang-format` inside in each of these dirs.

CONTRIBUTING.rst
```diff
@@ -68,6 +68,30 @@ We use the `Fork & Pull Request workflow`_ to ensure everything works smoothly.
...
+#. Create a ``build`` directory and call ``cmake`` from it (for now check the `.travis.yml` file for an example on CMake parameters), OR, on Windows, call the ``inv msvc`` task to generate a project under ``build\msvc`` directory, open it in the IDE and build the ``INSTALL`` project. (``inv`` is short for ``invoke``, from the `Invoke <https://www.pyinvoke.org/>`_ tool.)
+
+We follow some standards for the sake of code style, uniformity and quality. Don't worry, everything is provided through
+``pre-commit``. If you choose to develop Reaktoro using Conda, then the steps depicted above include ``pre-commit``.
+If you installed the dependecies manually, you have to install ``pre-commit`` just after cloning Reaktoro's
+repository. Just do in your console: ``pre-commit install``, then checkers and formatters are enable to verify every
+modification you introduced in a commit automatically. If something is not according to the standards, then the commit
+will fail, but ``pre-commit`` will modify the necessary part in order to fit it in our standards, then just commit
+again and everything will be fine.
+
```
volpatto: I added this part to CONTRIBUTING guidelines file. This replicates the Quick-Start in README, but provides additional information and instructions about `pre-commit`.

README.rst
```diff
@@ -151,6 +151,7 @@ Reaktoro using Conda, these are the steps:
 #. Install ``conda-devenv``: ``conda install -n base conda-devenv``
 #. Create an environment for Reaktoro, from the repository root directory: ``conda devenv``
 #. Activate the environment: ``source activate reaktoro`` from Linux/macOS or ``activate reaktoro`` from Windows
+#. Install pre-commit in order to activate git hooks, checkers and formatters: ``pre-commit install``
```
volpatto: I think that we have to warn new developers about `pre-commit`, since it will run in CI.

volpatto:
I think that I'm done. This PR is ready to be fully reviewed.

volpatto:
@allanleal, I put your `.clang-format` configs. Now, we are using `clang-format` from `clang-tools` installed through `conda`. This sounds somewhat like overkilling, but this way we can assure multiplatform development safer, plus the fact that we can control which `clang-format` is used in Reaktoro for everyone. 

Moreover, delegating to devs the responsibility of installing `clang-format` increases the complexity and instructions in order to build a Reaktoro development environment. The price we pay? Not a cheap one. This way we add around 100 MB to Reaktoro env. But, in my opinion, this is worthy. What do you think (or eventually any Reakoto's dev who's reading this and want to give an opinion)?

README.rst
volpatto: Oh, actually `pre-commit` is already installed within `conda`. It's listed as a dependency in `environment.devenv.yml`. My bad!

tadeu:
> 
> 
> Just to emphasize, I need help with `clang-format` inside `pre-commit`. I never did it before and I'm struggling with it. Could you please help me, @tadeu?

Sorry, I was on a trip and could only check this now, but it looks like it's all solved already, right? ;)

volpatto:
> > Just to emphasize, I need help with `clang-format` inside `pre-commit`. I never did it before and I'm struggling with it. Could you please help me, @tadeu?
> 
> Sorry, I was on a trip and could only check this now, but it looks like it's all solved already, right? ;)

Yes, I managed to get `clang-format` working! I think that I done here, btw.

tadeu:
There are a lot of commits here, and I didn't check all of them, but for code history sake, it's better to keep all formatting changes in just a single commit. You may try to rebase commits that change configuration, and then commit all formatting changes together, also using a dummy author name such as using `git commit --author "Auto-formatter <reaktoro@reaktoro.org>"`. Also, take a look here:

https://twitter.com/llanga/status/1163767833706323970

You can create a file named `.git-blame-ignore-revs` listing the commit hash where things were auto-formatted, so that this can be used to skip these changes when checking git history.

allanleal:
Hi Tadeu, yes, it was a simple fix on the GitHub branch permissions.
Thanks, no worries!

>

volpatto:
> There are a lot of commits here, and I didn't check all of them, but for code history sake, it's better to keep all formatting changes in just a single commit. You may try to rebase commits that change configuration, and then commit all formatting changes together, also using a dummy author name such as using `git commit --author "Auto-formatter <reaktoro@reaktoro.org>"`. 

Sure, the diff is very hard to track due to formatting changes.

> Also, take a look here:
> 
> https://twitter.com/llanga/status/1163767833706323970
> 
> You can create a file named `.git-blame-ignore-revs` listing the commit hash where things were auto-formatted, so that this can be used to skip these changes when checking git history.

Great, I didn't know about that. Nice tip, I will try to apply it here.

Reaktoro/Common/ChemicalVector.hpp
```diff
@@ -68,38 +68,40 @@ class ChemicalVectorBase
     /// Construct a ChemicalVectorBase instance with number of rows equal to given number of species.
     /// @param nspecies The number of nspecies in the chemical vector.
     explicit ChemicalVectorBase(Index nspecies)
-    : ChemicalVectorBase(nspecies, nspecies) {}
+        : ChemicalVectorBase(nspecies, nspecies) {}
```
allanleal, line 71: Best if your IDE do not format the files on save. This is generating too many changes.

allanleal: Best if your IDE does not format the files on save. This is generating too many changes.

volpatto: I think we should drop this PR. Maybe something can be implemented in this sense for future Reaktoro versions.

PR #136, opened 2020-02-13
Add openlibm support

tadeu:


tadeu:
Hey @allanleal, just to let you know that I've added an optional `openlibm` support and the jobs are passing (I've changed Python 3.6 jobs to use it, Python 3.7 jobs still use the default libm from each system).

If you want to try, there are only two things needed: setting the `REAKTORO_USE_OPENLIBM` environment variable so that `conda devenv` will also bring the `openlibm` dependency, and passing `-DREAKTORO_USE_OPENLIBM=1` to `cmake` (`inv compile` will do this automatically if the environment variable is defined).

Let me know if you are ok with it, then I'll add some documentation and changelog.

tadeu:
I've measure to see if there's any performance difference on Windows when running the tests, but didn't notice anything significant:

With openlibm: 17.24 s
Without openlibm: 16.89 s

The difference is of ~2%, which might well be on par with the fluctuations in time measurements.

allanleal:
Hi @tadeu - thanks for this. Sorry for the delay. I'll try to review this later today or early tomorrow morning.

tests/conftest.py
```diff
@@ -287,14 +288,21 @@ def equilibrium_inverse_with_h2o_nacl_caco3_co2_calcite_fixed_phase_volume():
...
+def _get_basename(request):
+    if is_using_openlibm():
+        import re
+        return re.sub(r"[\W]", "_", request.node.name) + ".openlibm"
+    return None  # use default
+
+
 @pytest.fixture(scope="function")
-def state_regression(num_regression):
+def state_regression(num_regression, request):
```
allanleal: Is the `request` argument a new addition to `pytest-regressions`? Just curious.

tadeu:
> 
> 
> @tadeu Interesting, so, `std:sqrt`, `std::exp`, etc., how do they get transformed into something else? I see no changes in source and header files.

Yes, the linker does the magic, `libopenlibm.a` has the same name for all these functions (`sqrt`, `exp`), so they end up replacing these functions from system's `libm`.

tests/conftest.py
tadeu: Ahh, this one is a fixture from plain `pytest`: https://docs.pytest.org/en/latest/reference.html#request

It's very useful for getting information like the test name!

allanleal:
Hi @tadeu - you mentioned something about postponing the merge of this PR. I see you made changes yesterday. Shall we merge it now?

cmake/Findopenlibm.cmake
```diff
@@ -0,0 +1,69 @@
+# Finds the openlibm library
```
allanleal: Does openlibm support [CMake config files](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file)? If so, we don't need this FindPackage module.

python/PyReaktoro/Common/PyIsUsingOpenlibm.cpp
```diff
@@ -0,0 +1,35 @@
+// Reaktoro is a unified framework for modeling chemically reactive systems.
```
allanleal: Can we name this file `PyOpenlibm.cpp`, in case we decide to add further exports in it in the future?

python/PyReaktoro/Common/PyIsUsingOpenlibm.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+bool is_using_openlibm() {
+#if defined(REAKTORO_USE_OPENLIBM) && REAKTORO_USE_OPENLIBM
+    return true;
+#else    
+    return false;
+#endif
+}
+
+void exportIsUsingOpenlibm(py::module& m)
```
allanleal: And then name this method `exportOpenlibm`.

tadeu:
The changes I made was just to solve conflicts with `master`, and yeah, we are good to merge it, please do the honors 😄

python/PyReaktoro/Common/PyIsUsingOpenlibm.cpp
```diff
@@ -0,0 +1,35 @@
...
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+#include <PyReaktoro/PyReaktoro.hpp>
+
+namespace Reaktoro {
+
+bool is_using_openlibm() {
```
allanleal: I know PEP8 is snake case, but for historical reasons, we have camelCase for methods in Reaktoro (originally C++). Can we name this as `usingOpenlibm` for consistency reasons?

cmake/Findopenlibm.cmake
tadeu: It is built with common `makefile` (and with `mingw` in Windows), so it doesn't output CMake config files. This module also adds the flags that are necessary for that "symbol substitution" that I talked about before (so that, e.g., `cos` really uses `cos` from `libopenlibm`, not from `libm`).

python/PyReaktoro/Common/PyIsUsingOpenlibm.cpp
tadeu: Sure, good idea!

- [ ] rename to `PyOpenlibm`

python/PyReaktoro/Common/PyIsUsingOpenlibm.cpp
tadeu: Ahh, sure!

- [ ] `is_using_openlibm` -> `isUsingOpenlibm`

tadeu:
I've applied the suggestions :+1:

allanleal:
Great - I'll merge once CI ends.

PR #139, opened 2020-03-01
Refactor devenv file to improve dependency handling

volpatto:
Hi all,

This PR does a refactor in the `conda-devenv` arrangement:

1. Some `sphinx` deps are available directly through `conda-forge`. Since it's better to have everything from the same place (in the case, `conda-forge`), I moved such deps from `pip` to `conda-forge`;

2. In order to have a better organization, I decomposed `environment.devenv.yml` in several small files, but with a clear purpose for each file. Please check it out at `.dependencies` dir. Now, in the `environment.devenv.yml`, we include files from `.dependencies`. This is an arguable solution, but I believe that this way is better for maintenance.

Feedbacks, reviews and suggestions are heavily welcome.

Cheers!

### DIFF
--- a/.dependencies/cpp.devenv.yml
+++ b/.dependencies/cpp.devenv.yml
@@ -1,0 +1,9 @@
+{% set debug = os.environ.get("CONFIG", "").lower() == "debug" %}
+
+dependencies:
+  - boost=1.70
+  - pybind11
+  - thermofun
+  {% if not debug %}
+  - pugixml                        # cannot link the Release version of `pugixml` (from conda) to the Debug version of Reaktoro
+  {% endif %}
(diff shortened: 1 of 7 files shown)


.dependencies/cpp.devenv.yml
```diff
@@ -0,0 +1,9 @@
+{% set debug = os.environ.get("CONFIG", "").lower() == "debug" %}
+
+dependencies:
+  - boost=1.70
+  - pybind11
+  - thermofun
```
volpatto: Is `thermofun` in the right place? I checked it and `thermofun` is a `cpp` package, but not a `cpp` tool for `reaktoro`.

volpatto:
See #140

PR #140, opened 2020-03-01
Refactor devenv file to improve dependency handling

volpatto:
Hi all,

This PR does a refactor in the `conda-devenv` arrangement:

1. Some `sphinx` deps are available directly through `conda-forge`. Since it's better to have everything from the same place (in the case, `conda-forge`), I moved such deps from `pip` to `conda-forge`;

2. In order to have a better organization, I decomposed `environment.devenv.yml` in several small files, but with a clear purpose for each file. Please check it out at `.dependencies` dir. Now, in the `environment.devenv.yml`, we include files from `.dependencies`. This is an arguable solution, but I believe that this way is better for maintenance.

Feedbacks, reviews and suggestions are heavily welcome.

Cheers!

### DIFF
--- a/.dependencies/docs.devenv.yml
+++ b/.dependencies/docs.devenv.yml
@@ -1,5 +1,4 @@
 dependencies:
-  # - doxygen=1.8.9                 # [linux]
   - graphviz                       # [linux]
   - sphinx==1.8.5
   - sphinx-autobuild
--- a/.dependencies/docs.devenv.yml
+++ b/.dependencies/docs.devenv.yml
@@ -1,0 +1,8 @@
+dependencies:
+  # - doxygen=1.8.9                 # [linux]
+  - graphviz                       # [linux]
+  - sphinx==1.8.5
+  - sphinx-autobuild
+  - sphinx_rtd_theme
+  - pip:
+    - sphinxcontrib-images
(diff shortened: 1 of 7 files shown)


.dependencies/docs.devenv.yml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  # - doxygen=1.8.9                 # [linux]
```
tadeu:
```suggestion
```

tadeu:
For me, I think that a single file with dependencies organized in blocks is slightly better than multiple files, because I like to have everything in the same view (organizing in multiple files is nice when we need to reuse dependencies for other environments). But, in practice, I find the difference to be kinda small, so either way is fine. ;)

volpatto:
> I think this looks nice and it simplifies the main devenv file, which I agree it was getting complicated.
> 
> @tadeu What do you think? I'll leave this approved, but it makes sense you review this before merging.

Nice, I'd prefer a separate arrangement because we do know exactly what kind of dependency each package is, and the visual is better. But this is a personal preference.

volpatto:
> For me, I think that a single file with dependencies organized in blocks is slightly better than multiple files, because I like to have everything in the same view (organizing in multiple files is nice when we need to reuse dependencies for other environments). But, in practice, I find the difference to be kinda small, so either way is fine. ;)

Yeah, I agree and understand you! I'm suggesting this refactor seeking a better visual aspect in our `environment.devenv.yml` file. In my opinion, it looks messy (and my environment was broke haha)

PR #141, opened 2020-03-02
Chemical output extension and refactor

volpatto:
Hi @allanleal and contributors,

This PR adds an extension to `ChemicalOutput` at Python interface. Let me elaborate it further:

1. I'm introducing a new format to `ChemicalOutput` from an kinetic path calculation, which is `ChemicalOutput.DataFrame()`. Given an output file specified to the kinetic path, this method reads and provide a `pandas.DataFrame` out-of-the-box. This is a good feature, because you can handle results easily with a `DataFrame`.

2. My second modification is a refactor. In `python/reaktoro` dir, the bindings from C++ are configured and properly construct, exposed and then installed. There, we already have extension for `ChemicalOutput`, one for the results as `numpy.ndarray` and another for `dict`. I simply refactored it, adding a new module dedicated only for `ChemicalOutput` extensions. I named it as `_chemicaloutput`. It's a private module, since we don't want that users find it. It's called by `__init__.py`, in order to load the extensions. I think that this is a good design, since it also provides a path for more extensions, if we need more.

That's all for now. I hope that this PR can be useful.

Cheers!


python/reaktoro/__init__.py
```diff
@@ -1,23 +1,2 @@
...
-
-# Define a method to convert the data in the file into a dictionary, with the 
-# headings being the keys, and the column data as the values.
-def _ChemicalOutput_dict(self):
-    data = numpy.loadtxt(self.filename(), skiprows=1)
-    columns = [ data[:, i] for i in xrange(data.shape[1]) ]
-    return { heading:column for heading,column in zip(self.headings(), columns) } 
-
-ChemicalOutput.dict = _ChemicalOutput_dict
+from reaktoro._chemicaloutput import *
```
allanleal: Can we have instead `from reaktoro.extensions.chemicaloutput import *`? We should then have this directory `extensions` under `rootdir/python/reaktoro` with Python extensions for possibly other classes too in the future.

python/reaktoro/_chemicaloutput.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
+import pandas as pd
+from reaktoro.PyReaktoro import ChemicalOutput
+
+
+def _to_array(self):
```
allanleal: When we do `from reaktoro._chemicaloutput import *`, or `from reaktoro.extensions.chemicaloutput import *` (see previous comment), is there a way to prevent importing these methods `_to_array_`, etc.? If we have other extension modules, and similar function names, could there be conflicts?

volpatto:
@tadeu would you mind to give us a review here as well?

TIA!

python/reaktoro/__init__.py
volpatto: Well, I definitely agree with you. But I failed when I tried it. My guess is the way we are generating Reaktoro bindings to Python. I didn't understand the CMake file, neither why we have a `setup.py.in` instead of a `setup.py` directly, but this is due my lack of knowledge. I hope @tadeu could explain it to me, please :)

python/reaktoro/_chemicaloutput.py
volpatto: Yeah, I think that it can give rise to name collisions, although no problem will be raised by Python interpreter, I guess. It will probably overwrite the function, and the last one imported will be the valid one in `reaktoro` namespace. However, for class extension, I think that the name used in the module scope will be used instead. But this is not an elegant solution, so we could improve it. Possible solution: more elaborated names, for instance, `_ChemicalOutput_to_array`.

volpatto: I will rename.

volpatto:
@allanleal and @tadeu: My first idea was to get ChemicalOutput stored values directly, but the quick solution was to retrieve them from output file. I'm open to suggestions in this regard.

python/reaktoro/__init__.py
tadeu: Probably just add to `packages=['reaktoro'],`, like `packages=['reaktoro', 'reaktoro.extensions'],`

python/reaktoro/_chemicaloutput.py
tadeu: When we import with the wildcard (`from ... import *`), the names that begin with `_` will not be automatically imported, but they can still be imported _explicitly_.

allanleal:
Looks good now. Let's follow then the convention `ClassName.new_method = _ClassName_new_method` for future class extension codes.

Last thing. Can we simplify/beautify file/directory names by removing the prefix underline? `reaktoro_extensions` instead of `reaktoro._extensions`? Also `chemicaloutput.py` instead of `_chemicaloutput.py`? Is there a compelling reason for this?

python/reaktoro/_extensions/chemicaloutput.py
```diff
@@ -0,0 +1,46 @@
...
+def _ChemicalOutput_to_dict(self):
+    """
+    Define a method to convert the ChemicalOutput data in the file into a dictionary,
+    with the headings being the keys, and the column data as the values.
+
+    :return:
+        A dict with ChemicalOutput properties from result file.
+    :rtype dict:
+    """
+    data = np.loadtxt(self.filename(), skiprows=1)
```
tadeu: Not for this PR, but a suggestion for later: it would be nice to allow `ChemicalOutput` in C++ to output to a string (or, even better, to an array), so that these conversions don't need to pass through the filesystem (outputting to a file just to read it later is very slow)

python/reaktoro/_extensions/chemicaloutput.py
```diff
@@ -0,0 +1,46 @@
...
+
+def _ChemicalOutput_to_array(self):
+    """
+    Define a method to convert the file data into an array.
+
+    :return:
+        An numpy array with data from ChemicalOutput.
+    :rtype numpy.ndarray:
+    """
+    output_array = lambda self: np.loadtxt(self.filename(), skiprows=1)
```
tadeu: `to_array` returns a lambda, but `to_dict` returns the dict itself.. is this really supposed to return a lambda here?

python/reaktoro/_extensions/chemicaloutput.py
```diff
@@ -0,0 +1,46 @@
...
+
+    :return:
+        A pandas DataFrame with ChemicalOutput data.
+    :rtype pd.DataFrame:
+    """
+    output_dataframe = pd.DataFrame(self.dict())
+    return output_dataframe
+
+
+ChemicalOutput.array = _ChemicalOutput_to_array
```
tadeu: Could these be renamed to `ChemicalOutput.to_array`, `ChemicalOutput.to_dict`, `ChemicalOutput.to_data_frame`? It's more in accordance with the many `to_*` methods used in other libraries such as `pandas`. (Some libraries also use `as_`, as in `ChemicalOutput.as_array`, etc., it could be an option)

python/reaktoro/_extensions/chemicaloutput.py
tadeu: Minor, but I think that `_ChemicalOutput_to_array` could be reused inside here

tests/test_chemicaloutput_extension.py
```diff
@@ -0,0 +1,115 @@
...
+    output = path.output()
+    output.filename(output_file_name)
+    for property_name, unit in dict_with_properties_to_output.items():
+        if property_name == "time(units=minute)" or property_name == "pH":
+            output.add(property_name)
+        else:
+            output.add(unit, property_name)
+
+    return output
+
```
tadeu: I think that a test for `to_array` is missing

tests/test_chemicaloutput_extension.py
```diff
@@ -0,0 +1,115 @@
+import os
+
+import pandas as pd
+import pytest
+from reaktoro import ChemicalEditor, ChemicalSystem, ReactionSystem, Partition
+from reaktoro import EquilibriumProblem, equilibrate, KineticPath, ChemicalOutput
+
+
+@pytest.fixture()
+def output_file_name():
```
tadeu: Use the `tmpdir` fixture here (or `tmp_path`), tests shouldn't change any non-temporary file!

python/reaktoro/_extensions/chemicaloutput.py
volpatto: Perfect. This was my first idea. But I needed a quick solution to handle output data easily. I will modify `ChemicalOutput` in the future directly in C++.

python/reaktoro/_extensions/chemicaloutput.py
volpatto: I don't think so... thanks, I will improve it.

python/reaktoro/_extensions/chemicaloutput.py
volpatto: Yeah, I agree. Changing it right now, thanks.

python/reaktoro/_extensions/chemicaloutput.py
volpatto: Fixed

tests/test_chemicaloutput_extension.py
volpatto: Fixed

tests/test_chemicaloutput_extension.py
volpatto: Nice, working on it

volpatto:
> Looks good now.
> 
> Let's follow then the convention `ClassName.new_method = _ClassName_new_method` for future class extension codes.

Yeah, I think this is a good pattern to follow.

> Last thing. Can we simplify/beautify file/directory names by removing the prefix underscore character `_`?
> 
> `reaktoro.extensions` instead of `reaktoro._extensions`, and `chemicaloutput.py` instead of `_chemicaloutput.py`?
> 
> Is there a compelling reason for having this leading underscore character?

Yes, the rationale behind them is to avoid direct usage of "private" pieces of code. Putting an underscore makes `_extension` as a hidden module. To use it, the user must know beforehand its existence. I did it this way because the extension is on a feature/object (`ChemicalOutput`) implemented originally in the C++ layer. Of course, everything is arguable here. You and @tadeu can freely give me the directions to follow. What do you guys think? For `chemicaloutput.py`, since it's inside `reaktoro._extensions`, there is no need to put an underscore in its beginning (`_extensions` is private already), so I changed it as you suggested!

volpatto:
Sorry, guys! I replied a comment and clicked in "Close and comment" haha My bad.

tadeu:
> Yes, the rationale behind them is to avoid direct usage of "private" pieces of code. Putting an underscore makes `_extension` as a hidden module. To use it, the user must know beforehand its existence. I did it this way because the extension is on a feature/object (`ChemicalOutput`) implemented originally in the C++ layer. Of course, everything is arguable here. You and @tadeu can freely give me the directions to follow. What do you guys think? For `chemicaloutput.py`, since it's inside `reaktoro._extensions`, there is no need to put an underscore in its beginning (`_extensions` is private already), so I changed it as you suggested!

Yeah, if it's not supposed to be imported by users (only internally), it's a convention to prefix the module with `_`.

tests/test_chemicaloutput_extension.py
volpatto: @tadeu I need assistance here. I pushed the modification with `tmp_path` fixture. However, `output.filename` is not working anymore. I don't know what is going on. Debug `reaktoro` through Python usage is very hard for me, unfortunately.

volpatto:
Okay, I finally fixed it. I did the requested changes, @tadeu. @allanleal, this is just a ping in case that you want to do further review.

tests/test_chemicaloutput_extension.py
```diff
@@ -0,0 +1,120 @@
...
+
+    path = KineticPath(reactions)
+    path.setPartition(partition)
+
+    return path, state
+
+
+@pytest.fixture
+def output_from_path(
+    brine_co2_path, tmp_path, dict_with_properties_to_output
```
tadeu: If you want, typing hints here could make it more clear that the two `path`'s refer to different things:

```python
def output_from_path(
    brine_co2_path: Tuple[KineticPath, ChemicalState],
    tmp_path: pathlib.Path,
    dict_with_properties_to_output,
):
```

tadeu:
Thanks @volpatto, LGTM!

allanleal:
Hi @volpatto - go ahead and merge please! Thanks for this, and thanks for the explanation of `_extensions` meaning/indicating a private module in Python!

tests/test_chemicaloutput_extension.py
volpatto: Great idea. It's indeed confusing.

PR #142, opened 2020-03-09
Add option to configure precision when outputting ChemicalState

tadeu:
This should allow better logging and reproducibility of simulations


python/PyReaktoro/Core/PyChemicalState.cpp
```diff
@@ -165,7 +168,8 @@ void exportChemicalState(py::module& m)
         .def("phaseAmount", phaseAmount4)
         .def("phaseStabilityIndices", &ChemicalState::phaseStabilityIndices)
         .def("properties", &ChemicalState::properties, py::keep_alive<1, 0>()) // keep returned ChemicalProperties object alive until ChemicalState object is garbage collected!
-        .def("output", &ChemicalState::output)
+        .def("output", output1, py::arg("out"), py::arg("precision") = 6)
+        .def("output", output2, py::arg("out"), py::arg("precision") = 6)
```
allanleal: Hi @tadeu - I added default precision here in the arg `precision`.

tadeu: Oh, good, thanks! I had to transcribe from a different commit I had in our fork (couldn't cherry pick because of conflicts), and I forgot this default here :)

allanleal:
Thanks @tadeu !

Issue #144, opened 2020-03-23
New conftest needs to be moved to tests dir


allanleal:
There is a new `conftest.py` file after merging PR #141 under `reaktoro/Reaktoro`, but this should be moved to `reaktoro/tests` (perhaps merged with the existing one `reaktoro/tests/conftest.py`). 

There is also regression data being stored under the C++ source dir:

~~~
reaktoro/Reaktoro/Core/ChemicalState-test/test_chemical_state_output_with_gaseous_phase_only.txt
~~~

which should be stored in `reaktoro/tests/regression`.

volpatto:
Solved in #146, so I'm closing this issue.

PR #146, opened 2020-04-13
Rename and move python tests and conftest.py to a proper dir

volpatto:
Hi all,

This PR aims to solve #144.

The modifications are the following:

* Several Python tests were scattered in Cpp source directories. I moved them to `tests` dir, which I think that is the correct dir.

* Old `conftest.py` inside Cpp source dir was merged with `conftest.py` inside `tests`. All fixtures are preserved then.

Cheers!


volpatto:
Well... this is failing due to some `ThermoFun` issue. I don't know how to solve it, but I noticed that it has a recent new release. I don't know it it's related though.

volpatto:
> Well... this is failing due to some `ThermoFun` issue. I don't know how to solve it, but I noticed that it has a recent new release. I don't know if it's related though.

Pinning `ThermoFun` does not solve the problem, unfortunately

tadeu:
@volpatto, I'll check the `ThermoFun` issue

volpatto:
> @volpatto, I'll check the `ThermoFun` issue in a different PR

Thanks! Fixed in #148

tests/conftest.py
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 
 import pytest
+import numpy as np
+
+from reaktoro import ChemicalEditor, ChemicalSystem, Partition
```
tadeu, line 6: It seems some of these are already imported in line 9, this line could be removed and `Partition` could be added there

volpatto: Ops! my bad

PR #148, opened 2020-04-14
Fix ThermoFun bindings issues

tadeu:
Some tests were failing because `pybind11` couldn't detect ThermoFun
objects being passed to Reaktoro functions. This was caused by
mismatched versions of pybind11 (2.4.3 in Reaktoro and 2.5.0 in
ThermoFun)


allanleal:
Thanks Tadeu!

PR #158, opened 2020-09-17
Improve openlibm use

allanleal:
This PR is a tentative solution to eliminate the need to update two sets of regression data: one using `openlibm`, and another not using it.

This PR forces `REAKTORO_USE_OPENLIBM=1` on the CI providers (currently AppVeyor and Travis).

Thus, any future PR in which **regression data is updated** using `pytest --force-regen` has to obey the following rule: 

*Reaktoro has to be compiled locally using `REAKTORO_USE_OPENLIBM=ON`.*

If this is not done, the pushed updated regression data will not conform with the calculated values in the CI providers and a failure will happen.

@tadeu Let me know if there is any loophole in this logic! :)


.dependencies/cpp.devenv.yml
```diff
@@ -1,16 +1,12 @@
...
   # pybind11 version must match the version used to compile ThermoFun
   - pybind11 >=2.5.0
   - thermofun==0.3.3
   - nlohmann_json==3.6.1
-  {% if not debug %}
-  # cannot link the Release version of `pugixml` (from conda) to the Debug version of Reaktoro
-  - pugixml
-  {% endif %}
-  {% if use_openlibm %}
   - openlibm
```
allanleal, line 9: This is a small lib, so we always consider it in the dev environment.

.travis.yml
```diff
@@ -12,15 +12,15 @@ matrix:
       env: PY_VER=3.6 REAKTORO_USE_OPENLIBM=1
 
     - os: linux
-      env: PY_VER=3.7
+      env: PY_VER=3.7 REAKTORO_USE_OPENLIBM=1
```
allanleal, line 15: This is to force all jobs to use `openlibm`.

appveyor.yml
```diff
@@ -16,6 +16,7 @@ environment:
   # Testing only Debug + Python 3.6 and Release + Python 3.7 so that the number of jobs is still 2 instead of 4.
   - CONFIG: Debug
     PY_VER: 3.6
+    REAKTORO_USE_OPENLIBM: 1
```
allanleal, line 19: Same reasoning here for AppVeyor.

tests/conftest.py
```diff
@@ -11,7 +11,6 @@
     Partition,
     EquilibriumProblem,
     EquilibriumInverseProblem,
-    isUsingOpenlibm,
```
allanleal, line 14: We now only have one dataset of regression data, so there is no need anymore to have the `*.openlibm.csv` files.

appveyor.yml
```diff
@@ -16,6 +16,7 @@ environment:
...
     #Temporally disabling regression tests, which otherwise would cause timeout failure (> 1 hour build time)
     TEST_DIR: Reaktoro
   - CONFIG: Release
```
tadeu, line 22: This job could be removed, Python 3.6 Debug and Python 3.7 Release looks good already

tadeu (approved):
Looks good!

appveyor.yml
allanleal: OK! Thanks! I'll remove this extra job in another PR just to avoid another almost 2h of AppVeyor run.

Issue #162, opened 2020-10-14
Investigate the Flaky in Phreeqc output format


OtavioPellicano:
Sometimes when I run `Phreeqc::execute()` method, the output change a little (see the image below). I noticed that change is always in `Solution composition`.

Code example:
```
    std::string database_path = "W:/reaktoro_processing/DATABASE/phreeqc.dat";
    std::string input_path = "W:/reaktoro_processing/IN/IW.pqi";
    std::string output_path = "W:/reaktoro_processing/OUT/IW.pqo";
    
    Phreeqc phreeqc(database_path);
    phreeqc.execute(input_path, output_path);
```

IW.pqo fragment:

![image](https://user-images.githubusercontent.com/27929707/96025487-91f3f900-0e2b-11eb-995d-cd397afaf47c.png)

OtavioPellicano:
I believe I found the error! If your input files are encoded as ANSI, instead of UTF-8, sometimes that line appears. One thing strange is that even with all inputs as UTF-8, the output will be encoded as ANSI.

allanleal:
OK, great - I'll close this issue then. I think any fix for such behavior would be better addressed directly in PHREEQC, and not in Reaktoro.

PR #165, opened 2020-10-21
Fix phreeqc stream destruction

### LINKED ISSUE #161: Investigate why Phreeqc destructor is throwing an exception
I made a test to run PHREEQC  input script by file:
```
    std::string database_path = "W:/reaktoro_processing/DATABASE/phreeqc.dat";
    std::string input_path = "W:/reaktoro_processing/IN/IW.pqi";
    std::string output_path = "W:/reaktoro_processing/OUT/IW.pqo";
    
    Phreeqc phreeqc(database_path);
    phreeqc.execute(input_path, output_path);
```
After I ran it, the Phreeqc destructor threw an exception caused by a raw pointer (see the image below)

image

OtavioPellicano:
fixes #161


tests/IW
```diff
@@ -0,0 +1,2 @@
+         sim	          pH	          pe	        temp	         Alk	     Acetate	          Ca	          Mg	          Na	          Cl	          Fe	        S(6)	       S(-2)
```
OtavioPellicano, line 1: these IW files were genereted automatically. Is there a way to don't do this?

Reaktoro/Interfaces/PhreeqcUtils.cpp
```diff
@@ -62,8 +62,8 @@ auto execute(PHREEQC& phreeqc, std::string input, std::string output) -> void
     int errors = phreeqc.run_simulations();
     phreeqc.Get_phrq_io()->clear_istream();
 
-    // Delete the dynamically allocated stream objects
-    if(out != nullptr) delete out;
+    // Set nullptr to stream object (to avoid delete an already deleted pointer in Phreeqc destructor)
+    out = nullptr;
```
OtavioPellicano, line 66: I tried to use shared_ptr<>, but it was not enough to solve the problem (it would be a little more tricky then we thought), so I decided to set `out` to `nullptr`. Just remove those lines work fine as well

thirdparty/phreeqc/utilities.cpp
```diff
@@ -611,13 +611,13 @@ dup_print(const char *ptr, int emphasis)
 		for (i = 0; i < l; i++)
 			dash[i] = '-';
 		dash[i] = '\0';
-		output_msg(sformatf("%s\n%s\n%s\n\n", dash, save_in.c_str(), dash));
-		log_msg(sformatf("%s\n%s\n%s\n\n", dash, save_in.c_str(), dash));
+		output_msg(sformatf(u8"%s\n%s\n%s\n\n", dash, save_in.c_str(), dash));
```
OtavioPellicano, line 614: I changed to UTF-8 because the `pytest-file_regression` was not working with ANSI encode. As I reported in issue #162 , the output file is being saved with ANSI encode, even if inputs  file are UTF-8.

tests/IW
tadeu: I don't know about the generation, but they probably shouldn't be committed? Perhaps add them to .gitignore?

tests/regression/test_phreeqc_execute/test_phreeqc_execute.txt
```diff
@@ -0,0 +1,194 @@
...
+	Mg                5.165e-02   5.165e-02
+	Na                5.021e-01   5.021e-01
+	S(-2)             6.316e-03   6.316e-03
+	S(6)              2.551e-02   2.551e-02
+
+----------------------------Description of solution----------------------------
+
+                                       pH  =   6.884    
+                                       pe  =   4.000    
+      Specific Conductance (�S/cm,  25�C)  = 52878
```
tadeu, line 62: It looks like it's not encoded in UTF-8

thirdparty/phreeqc/utilities.cpp
tadeu: Hmm, `u8` won't make much difference here, it's the contents of `save_in` that should be encoded in UTF-8 already. The `u8` prefix will be useful only if the non-ASCII characters are directly inside the string, like `u8"°C"`, but in this case there are no non-ASCII characters _directly_ in the string.

tests/regression/test_phreeqc_execute/test_phreeqc_execute.txt
OtavioPellicano: I think now it is fixed

thirdparty/phreeqc/utilities.cpp
allanleal: We need to be careful here because this is actual PHREEQC code - not Reaktoro. So, next time we update PHREEQC, this will be gone. The best would be to make such correction directly in their repository (or request this change there).

tests/IW
allanleal: What about avoiding these output files altogether (which can potentially change based on the system, locale, the next version of PHREEQC, etc.)? We could just create a regression test based on the `ChemicalState` contents after executing the PHREEQC script file and creating the `ChemicalState` object containing the compositional state in the `Phreeqc` object. Would this suffice?

Reaktoro/Interfaces/PhreeqcUtils.cpp
allanleal: Would it be possible to have a test case that only checks the failure you were having before is gone (i.e., when performing `Phreeqc::execute` under some conditions)?

thirdparty/phreeqc/utilities.cpp
tadeu: Good point, I hadn't realized that! Perhaps we should just define `NO_UTF8_ENCODING` then?

tests/IW
OtavioPellicano: It sounds good, but I don´t know how to do that. could you please show me an example?

Reaktoro/Interfaces/PhreeqcUtils.cpp
OtavioPellicano: Actually I tried to do something like you said, for avoid by now the encoding error, but I did not realize how. Do you have any suggestion?

tests/IW
allanleal: I think we don't even need the `ChemicalState`. We can use the methods available in [`Phreeqc`](https://reaktoro.org/cpp/classReaktoro_1_1Phreeqc.html) class to get the species amounts, temperature, pressure, activities, etc.

OtavioPellicano: I tried to get species amount, but the `IW` file was created. 
I just change the last two lines:
```
    p.execute(input_path)

    num_regression.check({'species_amount' : p.speciesAmounts()})
```

OtavioPellicano: we can also do not use regression at all, if you wish. Something like this:
```
    p.execute(input_path)

    assert approx(p.temperature()) == 298.15
    assert approx(p.pressure()) == 101324.99997084374
    assert approx(p.numElements()) == 11
    assert approx(p.numSpecies()) == 44
    assert approx(p.numPhases()) == 1
```

allanleal: That is even better indeed.

allanleal (approved):
Great, looks good!

tadeu (approved):

PR #173, opened 2021-01-11
Fixes of `develop` branch

mtsveta:
1) Extension of the error control method to account for variation in pressure and temperature
2) Adding the T and P update of the estimated `state` to make sure that it is not overwritten by the reference T and P


demos/cpp/demo-mass-imbalance-test.cpp
```diff
@@ -0,0 +1,138 @@
+#include <Reaktoro/Reaktoro.hpp>
```
allanleal, line 1: Let's remove this demo file from git stage

PR #176, opened 2021-01-29
Allow more recent versions of nlohmann_json depedency

tadeu:


allanleal (approved):
This makes more sense indeed! :)

PR #177, opened 2021-01-29
Separate CubicEOS Python bindings in a proper file and expose BIPs

volpatto:
This PR contains the following changes:

- [X] Move `CubicEOS::Params` definitions and bindings inside `PyFluidPhase` to a proper file `PyCubicEos`;
- [X] Expose setup Binary Interaction Parameters through Cubic EOS Python bindings;
- [X] Tests in Python for the exposed features. Please see `tests/test_cubiceos.py`;
- [X] Fix Database parser for `Liquid` type species;
- [X] Add a custom xml database for hydrocarbons components. Only for tests purpose;
- [x] Add a case solving a hydrocarbon Liquid-Vapor equilibrium using non-zero BIPs (k).


python/PyReaktoro/Thermodynamics/Models/PyCubicEos.cpp
```diff
@@ -0,0 +1,52 @@
+// Reaktoro is a unified framework for modeling chemically reactive systems.
```
allanleal, line 1: `CubicEos` -> `CubicEOS`

python/PyReaktoro/Thermodynamics/Models/PyCubicEos.cpp
```diff
@@ -0,0 +1,52 @@
...
+    py::class_<CubicEOS::Params>(m, "CubicEOSParams")
+        .def(
+            py::init<CubicEOS::Model, PhaseIdentificationMethod>(),
+            py::arg("model") = CubicEOS::PengRobinson,
+            py::arg("phase_identification_method") = PhaseIdentificationMethod::None
+        )
+        .def_readwrite("model", &CubicEOS::Params::model)
+        .def_readwrite("phase_identification_method", &CubicEOS::Params::phase_identification_method)
+        ;
+}
```
allanleal, line 50: Are you going to create bindings for `CubicEOSInteractionParamsArgs` and `CubicEOS::setInteractionParamsFunction` in this PR? I this what you needed?

python/PyReaktoro/Thermodynamics/Models/PyCubicEos.cpp
volpatto: Whoops, thanks!

python/PyReaktoro/Thermodynamics/Models/PyCubicEos.cpp
volpatto: Yes, we need to set such values for Cubic EOS models.

python/PyReaktoro/Thermodynamics/Models/PyCubicEos.cpp
volpatto: Fixed now, I think.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
```diff
@@ -248,10 +248,9 @@ struct CubicEOS::Impl
 
         // Calculate the table of binary interaction parameters and its temperature derivatives
         InteractionParamsResult kres;
-        InteractionParamsArgs kargs{T, a, aT, aTT, b};
```
volpatto, line 251: I think the common arg for BIPs is the temperature only. Most of the time BIPs have constant value or given as a polynomial, so I removed those other args.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
```diff
@@ -262,9 +261,9 @@ struct CubicEOS::Impl
         {
             for(unsigned j = 0; j < nspecies; ++j)
             {
-                const ThermoScalar r = kres.k.empty() ? ThermoScalar(1.0) : 1.0 - kres.k[i][j];
-                const ThermoScalar rT = kres.kT.empty() ? ThermoScalar(0.0) : -kres.kT[i][j];
-                const ThermoScalar rTT = kres.kTT.empty() ? ThermoScalar(0.0) : -kres.kTT[i][j];
+                const double r = kres.k.size() ? 1.0 - kres.k(i, j) : 1.0;
+                const double rT = kres.kT.size() ? -kres.kT(i, j) : 0.0;
+                const double rTT = kres.kTT.size() ? -kres.kTT(i, j) : 0.0;
```
volpatto, line 266: I have had to change this part since the bindings are straightforward using Eigen. The logic is the same. If a `MatrixXd` is just declared, the `size` would be `col * rows`, with `col = rows = 0`. This `k.size() = 0` if the BIPs are not defined.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
```diff
@@ -45,6 +46,24 @@ class CubicEOS
         VanDerWaals, RedlichKwong, SoaveRedlichKwong, PengRobinson,
     };
 
+    /// Class to define or store Binary Interaction Parameters (BIPs) from a calculation or input.
+    /// Note that the BIPs can depend on the temperature, thus kT and kTT should be also provided.
+    struct InteractionParamsResult
+    {
+        /// The BIPs matrix. The size must be (n, n), where n is the number of species
+        MatrixXd k;
```
volpatto, line 54: Despite the name, those `MatrixXd`s are aliases defined in `Reaktoro/Math/Matrix.hpp`.

python/PyReaktoro/Thermodynamics/Models/PyCubicEOS.cpp
```diff
@@ -0,0 +1,66 @@
...
+        .def_readwrite("phase_identification_method", &CubicEOS::Params::phase_identification_method)
+        .def_readwrite("binary_interaction_values", &CubicEOS::Params::binary_interaction_values)
+        ;
+
+    py::class_<CubicEOS::InteractionParamsResult>(m, "BinaryInteractionParams")
+        .def(
+            py::init<MatrixXd, MatrixXd, MatrixXd>(),
+            py::arg("k") = MatrixXd{},
+            py::arg("kT") = MatrixXd{},
+            py::arg("kTT") = MatrixXd{}
```
volpatto, line 58: I did this because it allows the user to define only `k` without defining `kT` and `kTT`. This is useful since most of the time `k` has constant entries. I don't think it has any side effect due to those [ternary operators](https://github.com/reaktoro/reaktoro/pull/177/files#r568251361).

volpatto: Btw, it is not a Ref MatrixXd. The reason is that if the user passes a `List` in Python (as in the tests below), it would raise an error due to type incompatibility.

volpatto:
No tests are provided to check the effect of BIPs on Equilibrium calculations in this PR. I'm planning to do it as soon as we merge this PR. I'm working on Hydrocarbon mixtures using Cubic EOS in [this branch](https://github.com/volpatto/reaktoro/tree/add-hydrocarbon-tests).

tests/test_cubiceos.py
```diff
@@ -0,0 +1,126 @@
...
+        bips_values = BinaryInteractionParams(k)
+        return bips_values
+
+    T_dummy = 2.0
+    bips_expected = bips_function(T_dummy)
+
+    cubic_eos_params = CubicEOSParams(binary_interaction_values=bips_function)
+    bips_calculated = cubic_eos_params.binary_interaction_values(T_dummy)
+    
+    assert np.allclose(bips_calculated.k, bips_expected.k)
```
tadeu, line 126: Could check for equality here, also checking that `kT` and `kTT` are zero

tadeu (changes requested):
Nice improvement!

tests/test_cubiceos.py
```diff
@@ -0,0 +1,126 @@
+import numpy as np
```
tadeu, line 1: About the file name, are we using the casing like `test_cubiceos.py` or like `test_CubicEOS.py`?

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
volpatto: I think it would be interesting to have a "size check" for those BIPs matrix. I mean, suppose we have `n` species. So all BIPs matrices have dimension `(n, n)`. I need to elaborate it further in the code.

tests/test_cubiceos.py
```diff
@@ -0,0 +1,126 @@
...
+        [0.00, 0.00, 0.00]
+    ]
+    kTT = [
+        [0.00, 0.00, 0.00],
+        [0.00, 0.00, 0.00],
+        [0.00, 0.00, 0.00]
+    ]
+
+    bips = BinaryInteractionParams(k, kT, kTT)
+    assert np.allclose(bips.k, k)
```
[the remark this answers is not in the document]
volpatto: Thanks! I will improve it as suggested

tests/test_cubiceos.py
volpatto: This is a good question. I would prefer `test_CubicEOS.py`.

tadeu: Just checked here, for consistency it should be lowercase (as it already is), I think that they were all renamed at some point :)

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
```diff
@@ -576,6 +579,41 @@ auto CubicEOS::setInteractionParamsFunction(const InteractionParamsFunction& fun
     pimpl->calculate_interaction_params = func;
 }
 
+auto CubicEOS::sanityCheckInteractionParamsFunction(const InteractionParamsFunction& func) -> void
```
volpatto, line 582: I added this function to check if BIPs have the right dimensions.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
```diff
@@ -498,6 +497,10 @@ CubicEOS::CubicEOS(unsigned nspecies, CubicEOS::Params params)
 {
     pimpl->model = params.model;
     pimpl->phase_identification_method = params.phase_identification_method;
+    if(params.binary_interaction_values) {
+        sanityCheckInteractionParamsFunction(params.binary_interaction_values);
```
volpatto, line 501: Now we have a sanity check for BIPs. Please [see here](https://github.com/reaktoro/reaktoro/pull/177/files#r568849826).

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
volpatto: > I think it would be interesting to have a "size check" for those BIPs matrix. I mean, suppose we have `n` species. So all BIPs matrices have dimension `(n, n)`. I need to elaborate it further in the code.

Implemented [here](https://github.com/reaktoro/reaktoro/pull/177/files#r568849826).

Reaktoro/Thermodynamics/Core/Database.cpp
```diff
@@ -752,6 +752,17 @@ struct Database::Impl
...
+            {
+                LiquidSpecies liquid_species = parseFluidSpecies(node);
+                liquid_species.setName(name);
+                const auto liquid_species_suffix_size = 3;
+                liquid_species.setName(name.substr(0, name.size() - liquid_species_suffix_size) + "(liq)");
+                if(valid(liquid_species))
+                {
+                    liquid_species_map[liquid_species.name()] = liquid_species;
+                }
+            }
```
volpatto, line 765: This modification was necessary to properly read hydrocarbon data from the custom xml database.

tests/data/hydrocarbons.xml
```diff
@@ -0,0 +1,159 @@
+<?xml version="1.0" ?>
```
volpatto, line 1: A custom Database containing Hydrocarbon data to be used in Liquid phases. This is only for the tests.

volpatto:
> This PR contains the following changes:
> 
> * [x]  Move `CubicEOS::Params` definitions and bindings inside `PyFluidPhase` to a proper file `PyCubicEos`;
> * [x]  Expose setup Binary Interaction Parameters through Cubic EOS Python bindings;
> * [x]  Tests in Python for the exposed features. Please see `tests/test_cubiceos.py`;
> * [x]  Fix Database parser for `Liquid` type species;
> * [x]  Add a custom xml database for hydrocarbons components. Only for tests purpose.

The PR is now tackling an increased number of issues. Should I rename the PR? Any suggestions?
I'm planning to add a hydrocarbon case using non-zero BIPs and stop at this point.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
```diff
@@ -262,9 +261,9 @@ struct CubicEOS::Impl
         {
             for(unsigned j = 0; j < nspecies; ++j)
             {
-                const ThermoScalar r = kres.k.empty() ? ThermoScalar(1.0) : 1.0 - kres.k[i][j];
-                const ThermoScalar rT = kres.kT.empty() ? ThermoScalar(0.0) : -kres.kT[i][j];
-                const ThermoScalar rTT = kres.kTT.empty() ? ThermoScalar(0.0) : -kres.kTT[i][j];
+                const double r = kres.k.size() ? 1.0 - kres.k(i, j) : 1.0;
```
allanleal, line 264: If `k` has any dependency on temperature, this change here (using `double`) will prevent temperature derivatives.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
```diff
@@ -45,6 +46,24 @@ class CubicEOS
...
+        /// The derivative of each k entry w.r.t T.
+        MatrixXd kT;
+
+        /// The derivative of each kT entry w.r.t T.
+        MatrixXd kTT;
+    };
+
+    /// Function wrapper to calculate (temperature-dependent) binary interaction parameters.
+    using InteractionParamsFunction =
+        std::function<InteractionParamsResult(const double&)>;
```
allanleal, line 65: The argument here should be `const ThermoScalar&` (instead of `const double&`) which will be temperature. Otherwise, we are not able to compute temperature derivatives. Even for `kTT`, we are able to get `kTT.ddT` (with automatic differentiation).

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
```diff
@@ -153,10 +153,17 @@ class CubicEOS
...
+    /// Set the function that calculates the interaction parameters kij (k) 
+    /// and its temperature derivatives (kT and kTT).
+    /// @param func A wrapper function to calculate the BIPs matrices
     /// @see InteractionParamFunction, InteractionParamArgs, InteractionParamResult
     auto setInteractionParamsFunction(const InteractionParamsFunction& func) -> void;
 
+    /// Sanity check to verify if BIPs matrices have proper dimensions. Considering that the phase has
+    /// n species, the BIP matricies k, kT and kTT should have (n, n) as dimensions.
+    /// @see setInteractionParamsFunction
+    auto sanityCheckInteractionParamsFunction(const InteractionParamsFunction& func) -> void;
```
allanleal, line 165: This method declaration should not be exposed here - move it to `CubicEOS.cpp` (possibly as a free function), where it is used.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
allanleal: This will become `MatrixXr` in the v2.0 (matrix of `real` numbers, where `real` is an `autodiff` type)

tests/test_cubiceos.py
```diff
@@ -0,0 +1,150 @@
...
+        [0.00, 0.00, 0.00]
+    ]
+    kTT = [
+        [0.00, 0.00, 0.00],
+        [0.00, 0.00, 0.00],
+        [0.00, 0.00, 0.00]
+    ]
+
+    bips = BinaryInteractionParams(k, kT, kTT)
+    assert bips.k.all() == np.array(k).all()
```
tadeu, line 86: Should be:

```
assert np.all(bips.k == k)
```

or, for better error messages:

```
assert_array_equal(bips.k, k)
```

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
volpatto: Hmm.. I see! Well, it means that we should have `ThermoScalar` with Python bindings. I looked for it and I have not found it, maybe I'm missing something. Do you have any advice? `ThermoScalar` is pretty big, so it would demand some time to expose.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
volpatto: Sure, thanks

volpatto: Done.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
volpatto: Sorry, nevermind. Found everything I need in `PyAutoDiff.cpp`.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
volpatto: Done. I wonder if this is the better approach. What do you think about the code below?
```c++
const ThermoScalar r = kres.k.size() ? ThermoScalar(1.0 - kres.k(i, j)) : ThermoScalar(1.0);
```

I'm asking because I have changed the BIPs input using `ThermoScalar`, but with Eigen things are far simpler, IMHO.

Reaktoro/Thermodynamics/EOS/CubicEOS.hpp
volpatto: Done.

docs/sphinx/acknowledgements.rst
```diff
@@ -85,8 +85,9 @@ of modern software engineering practices into the project. We thank them below:
 * Edson Tadeu M. Manoel *(ESSS, Brazil)* *(2018-present)*
 * Alexandre Barbosa *(ESSS, Brazil)* *(2018-present)*
 * Bruno Oliveira *(ESSS, Brazil)* *(2018-present)*
-* Armando A. S. C. Neto *(ESSS, Brazil)* *(2018-present)*
+* Armando A. S. C. Neto *(ESSS, Brazil)* *(2018-2020)*
 * Svetlana Kyas, Dr. *(ETH Zurich, Switzerland)* *(2018-present)*
+* Diego T. Volpatto *(ESSS, Brazil)* *(2020-present)*
```
volpatto, line 90: I updated the acknowledgements. Please correct me if something is wrong.

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
allanleal: This is fine. It could also be:

~~~cpp
const auto r = kres.k.size() ? 1.0 - kres.k(i, j) : ThermoScalar(1.0);
~~~

The `ThermoScalar(1.0)` is needed otherwise you have a ternary operator that has two different return types (`ThermoScalar` from the `1 - kij` branch, and `double` from the `1.0` branch). 

This is simpler with `autodiff::real` in the v2.0.

docs/sphinx/acknowledgements.rst
allanleal: Looks good.

tests/test_cubiceos.py
```diff
@@ -65,26 +66,66 @@ def test_bips_setup():
...
     kTT = [
-        [0.00, 0.00, 0.00],
-        [0.00, 0.00, 0.00],
-        [0.00, 0.00, 0.00]
+        [kTT_00, kTT_01, kTT_02],
+        [kTT_10, kTT_11, kTT_12],
+        [kTT_20, kTT_21, kTT_22]
     ]
 
     bips = BinaryInteractionParams(k, kT, kTT)
```
volpatto: @allanleal I don't know if this is according to your expectations, but I set the BIPs to use `ThermoScalar` in the input. IMHO, the complexity drastically increased, but maybe this is the price to enable automatic differentiation. Please note below that now we can't handle BIPs matrices in Python directly as `List`s or `numpy.ndarray`s without a post-processing for-loop, just like the `assert`s in the final of this test.

Is it OK? Are we good to go?

Reaktoro/Thermodynamics/EOS/CubicEOS.cpp
volpatto: But what about the automatic differentiation? Some BIPs can depend on temperature... I'm trying to find a solution that keeps the simplicity of `Eigen::MatrixXd` as interface with the `ThermoScalar` auto diff capabilities.

Reaktoro/Thermodynamics/Core/Database.cpp
allanleal: Can we use `(l)` here without breaking other codes? This would be more preferable.

tests/test_cubiceos.py
volpatto: The ideal solution, IMHO, would be the implementation of `ThermoMatrix` based on `ThermoVector`. It was a little bit cumbersome to write each BIPs matrix input as `ThermoScalar`, but I think we can go with that as well.

volpatto: > The ideal solution, IMHO, would be the implementation of `ThermoMatrix` based on `ThermoVector`. It was a little bit cumbersome to write each BIPs matrix input as `ThermoScalar`, but I think we can go with that as well.

However, this would result in a considerable amount of work for me haha

Reaktoro/Thermodynamics/Core/Database.cpp
volpatto: Sure, we could define "Reaktoro's standard". I think we need to inform the user about "custom databases" somewhere, but no idea where. Any recommendation in this regard?

tests/test_cubiceos.py
allanleal: I'm getting rid of all these ThermoScalar, ThermoVector, ChemicalScalar, ChemicalVector in the v2.0 - we can have Eigen Matrix of autodiff::real there and this is what you want (but it's not worth doing similar things here, implementing a ThermoMatrix).

allanleal (approved):
Perfect! Thanks @volpatto .

Issue #179, opened 2021-02-02
Installed include/cvode/cvode.h conflicts with sundials that installs cvode


yurivict:
Instead of bundling cvode you should depends on cvode from the sundials package.

Sundials packages are available on about every distro: https://repology.org/project/sundials/versions

allanleal:
Yes, this superbuild approach in Reaktoro started a few started years ago before `conda` became its dependency manager. cvode is available in conda-forge, which is probably from where we'll pick it for our day-to-day development.

yurivict:
I don't use conda. Package managers don't use conda.

allanleal:
Conda is just to simplify dependency management. Not every library is available via package managers of operating systems.

You don't need to use conda if you are happy installing every dependency yourself. CMake `find_package` should pick up the libraries needed for you. 

At the moment, however, there is no `find_package(cvode)` in Reaktoro. We are bound to use the cvode library built together with Reaktoro.

yurivict:
> You don't need to use conda if you are happy installing every dependency yourself.

I would like to create a package for ```reaktoro``` and conda is not an option for package managers. conda downloads files, and all downloads need to be fingerprinted for security reasons, this is why conda is not compatible with package managers.

You should support both conda and non-conda users.

It should be possible to individually choose libraries that are used as packages, or as bundled dependencies.

allanleal:
Right. I'm open to alternatives. Let me know what you have in mind more specifically.

yurivict:
Please add cmake options for dependencies, for example ```USE_EXTERNAL_CVODE```.

allanleal:
Working on it. Just installed cvode v2.8.2 (I know, it's outdated) using:

~~~
mkdir build
cmake .. -DEXAMPLES_ENABLE=OFF -DEXAMPLES_INSTALL=OFF -DCMAKE_INSTALL_PREFIX=install
make -j
~~~

I'll now try to get `find_package(cvode)` to work and see if my changes can be committed.

allanleal:
I've create a PR for this. You can also check out the branch:

https://github.com/allanleal/reaktoro/tree/add-cmake-options-to-prevent-deps-from-building-automatically

We'll soon have version 2.0 of Reaktoro. In this new version, there are some new deps (and some removed). 

We would need packages for:

* https://github.com/autodiff/autodiff
* https://github.com/reaktoro/optima

[...]

PR #180, opened 2021-02-02
Enable the use of cvode from a given path using REAKTORO_PATH_TO_CVODE

allanleal:
This PR is a fix to issue #179 .

To use `cvode` from a given installation location, do:

~~~
mkdir build && cd build
cmake .. -DREAKTORO_PATH_TO_CVODE=/home/user/codes/cvode-2.8.2/build/install
make -j
~~~

Note: Do not add `/` to the end of `/home/user/codes/cvode-2.8.2/build/install`.

By defining the cmake variable `REAKTORO_PATH_TO_CVODE`, this will skip the automatic build of cvode and use the libraries `libsundials_cvode.so` and `libsundials_nvecserial.so` available in `path/to/install/lib`. The header files should be in `/path/to/install/include`.

### DIFF
--- a/Reaktoro/CMakeLists.txt
+++ b/Reaktoro/CMakeLists.txt
@@ -9,8 +9,15 @@ set(REAKTORO_SHARED_LIB ${PROJECT_NAME}${SUFFIX_SHARED_LIBS})
 set(REAKTORO_STATIC_LIB ${PROJECT_NAME}${SUFFIX_STATIC_LIBS})
 
 # Set the list of names of the third-party targets and libraries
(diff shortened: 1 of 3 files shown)


volpatto:
As a suggestion: maybe we should specify the CVODE version since Reaktoro internals heavily depends on an old CVODE API. If someone tries to link Reaktoro against an incompatible version of CVODE, I think it will break several parts in ODE solver. I have tried something like that before...

allanleal:
CVODE is not used anymore in Reaktoro v2, and it seems there is no urgent need to have this in Reaktoro v1 (where an old version of CVODE is used and already embedded in the source code).

PR #182, opened 2021-02-10
GitHub Actions as CI

volpatto:
This PR replaces Travis CI and AppVeyor with GitHub Actions. It replicates as close as possible the previous CI steps.


.github/workflows/windows.yml
```diff
@@ -0,0 +1,72 @@
...
+  schedule:
+    - cron: "0 5 * * 1"  # runs at 05:00 UTC on Mondays
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+
+    env:
+      use_openlibm: 1
```
tadeu, line 19: In Windows environment variables are case-insensitive, but just for the sake of consistency with Linux, perhaps using "all caps" here would be better (`USE_OPENLIBM` and `CONFIGURATION`).

By the way, this also could be directly `REAKTORO_USE_OPENLIBM`.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,71 @@
...
+    - cron: "0 5 * * 1"  # runs at 05:00 UTC on Mondays
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    env:
+      os: Linux-x86_64
+      use_openlibm: 1
```
tadeu, line 20: May be directly `REAKTORO_USE_OPENLIBM: 1`

.github/workflows/linux.yml
```diff
@@ -0,0 +1,71 @@
...
+  schedule:
+    - cron: "0 5 * * 1"  # runs at 05:00 UTC on Mondays
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    env:
+      os: Linux-x86_64
```
tadeu, line 19: Is this `os` environment variable required? It seems it's not used (the same for other files)

.github/workflows/linux.yml
```diff
@@ -0,0 +1,71 @@
...
+      use_openlibm: 1
+      NUMBER_OF_COMPILATION_JOBS: 2
+
+    strategy:
+      fail-fast: true
+      max-parallel: 2
```
tadeu, line 25: Out of curiosity, is there a special reason for this one to be 2 and the others 4?

tadeu (commented):
Awesome work @volpatto 🎉

.github/workflows/linux.yml
allanleal: @volpatto , we can use 4 jobs for all configs (linux, win, and osx). This is what I did here:
https://github.com/allanleal/reaktoro/tree/v2.0-preview/.github/workflows

```
max-parallel: 4
```

Just make sure that in Linux, the compilation uses 3 parallel jobs out of 4, otherwise, there is a tendency for the VM to freeze.

tasks.py
```diff
@@ -114,7 +114,7 @@ def msvc(c, clean=False, config='Release'):
             c,
             build_dir=build_dir,
             artifacts_dir=artifacts_dir,
-            cmake_generator="Visual Studio 15 2017",
+            cmake_generator="Visual Studio 16 2019",
```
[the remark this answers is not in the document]
volpatto: Ah, AFAIK, yes! Maybe there is an Actionto select MSVC, but natively no, only 2019, unfortunately.

tasks.py
```diff
@@ -136,11 +136,12 @@ def compile(c, clean=False, config='Release', number_of_jobs=-1, verbose=False):
         build_subdirectory="build",
     )
 
+    cmake_generator = "Ninja"
     cmake_command = _get_cmake_command(
         c,
         build_dir=build_dir,
         artifacts_dir=artifacts_dir,
-        cmake_generator="Ninja",
+        cmake_generator=cmake_generator,
```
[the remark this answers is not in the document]
volpatto: Yes, thanks! It's a leftover, I was defining `cmake_generator` by OS.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,72 @@
...
+          conda install conda-devenv
+          conda devenv
+      - name: Configuring Cache
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: .clcache
+          key: ${{ runner.os }}-${{ matrix.PY_VER }}-${{ github.ref }}
+      - name: Building and Installing Reaktoro
+        shell: powershell
```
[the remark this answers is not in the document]
volpatto: Yeah, I will adapt here

.github/workflows/windows.yml
volpatto: Great, I will do that. Thanks!

.github/workflows/linux.yml
volpatto: Yes, I have checked that with `pytest -n auto`. The number of parallel runs `pytest` uses is 2 or 3 (Linux is 2), never 4. Maybe hyper-threading is disabled in GH Actions VMs. This is why I have set those values. Should I still set everything as 4? Or should we go as it is now?

volpatto: Just an update: Linux and Win have 2 parallel runs with `pytest . -n auto`, while OSX has 3. This observation is in agreement with [GH Actions Docs](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners). It seems that maybe hyper-threading is really disabled or `pytest-xdist` is unable to capture the number of threads properly, IDK...

allanleal (approved):
Excellent Diego!

PR #189, opened 2021-02-23
Add publish step to Linux GH Actions workflow

volpatto:
This PR adds a step in the Linux GH Action that upload Reaktoro to PyPI automatically whenever a commit with a tag is pushed to master.


volpatto:
I will close this PR for now, since the secrets can not be read when triggering an action from a Fork.

PR #193, opened 2021-02-24
Set PyPI publish job with Github Actions

volpatto:


.dependencies/python.devenv.yml
```diff
@@ -7,4 +7,4 @@ dependencies:
   - tabulate
   - colorama
   - python={{ python_version }}
-  - twine
+  - setuptools<49.0
```
volpatto, line 10: Remark: for some reason, the package build with `setuptools>=49.0` breaks for some reason, maybe due to a new Python packaging standard. For now, this version constraint is enough.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,61 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - v*
+
+jobs:
+  publish:
```
volpatto, line 11: This is the job that publishes Reaktoro to PyPI. It runs only when a tagged commit is pushed to master.

.dependencies/python.devenv.yml
volpatto: This does not work for Windows. Pinning to 49.5.0.

ci/check_compiled_files.py
```diff
@@ -14,7 +14,13 @@
...
 
 if sys.platform.startswith('linux'):
-    assert (artifacts / 'lib/libReaktoro.so').is_file()
+    if (artifacts / 'lib64').is_dir():
+        assert (artifacts / 'lib64/libReaktoro.so').is_file()
+    elif (artifacts / 'lib').is_dir():
+        assert (artifacts / 'lib/libReaktoro.so').is_file()
+    else:
+        raise RuntimeError(f"Both {artifacts / 'lib64/libReaktoro.so'} " 
+            f"and {artifacts / 'lib/libReaktoro.so'} are not found.")
```
volpatto, line 23: CC @tadeu. For some reason, in my Manjaro I don't have `lib`, but I have `lib64` instead. This is a workaround that we can improve afterwards. I'm open for suggestions!

allanleal: I think in the conda-forge feedstock of autodiff, I had to set `CMAKE_LIBRARY_OUTPUT_DIRECTORY` to `lib` to ensure no `lib64` is used.

PR #194, opened 2021-02-25
Add a fix to Reaktoro install

volpatto:


python/reaktoro/CMakeLists.txt
```diff
@@ -15,7 +15,13 @@ configure_file(${SETUP_PY_IN} ${SETUP_PY})
...
     COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:PyReaktoro>
         ${CMAKE_CURRENT_BINARY_DIR}/$<TARGET_FILE_NAME:PyReaktoro>
-    COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} --quiet install --prefix=${CMAKE_BINARY_DIR}
+    COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} --quiet
+        sdist --dist-dir=${CMAKE_CURRENT_BINARY_DIR}/dist
+        bdist_wheel --dist-dir=${CMAKE_CURRENT_BINARY_DIR}/dist
+    COMMAND ${PYTHON_EXECUTABLE} -m pip install
+        reaktoro --prefix=${CMAKE_BINARY_DIR}
+        --find-links=\${CMAKE_CURRENT_BINARY_DIR}/dist
+        --no-index
```
volpatto, line 24: This `ALL` build has caused some problems in past builds. With the new Python install using `setuptools`, an upgrade in this code is required, as did above. Out of curiosity: @allanleal, since we already have `install` in this CMakeLists.txt, we we have to perform a Python install in the above custom target?

allanleal: The install in the `reaktoro` target is for the `build` operation (something done many times during development). The other install is for the `install` target. I think this is improved in the v2.0.

volpatto: Ah, I see! Well, fair enough.

PR #203, opened 2021-07-07
Add tests and improvements for equilibrium considering hydrocarbons

alexandrebbruno:
Add tests with hydrocarbons with code improvements

authored-by: Diego Volpatto <volpatto@lncc.br>
Co-authored-by: Alexandre B. Bruno <alexandrebbruno@gmail.com>


volpatto:
@allanleal this is the work I have done in the beginning of this year for Hydrocarbon mixtures. Unfortunately, I ran out of time and then @alexandrebbruno rescued this work (thanks a lot!) and made some improvements.

Note that the calculations inside `CubicEOS` are slightly different right now, and potentially can effect every mixture that uses this thermodynamic model.

This work is split in more than one PR. More PRs can be open in the future and there are rooms for improvements.

Reaktoro/Core/ChemicalProperties.cpp
```diff
@@ -321,6 +321,13 @@ auto ChemicalProperties::phaseSpecificVolumes() const -> ChemicalVector
     return phaseAmounts()/phaseMasses() % phaseMolarVolumes();
 }
 
+auto ChemicalProperties::phaseCompressibilityFactors() const -> ChemicalVector
```
allanleal, line 324: This class (named `ChemicalProps` in v2.0) should contain only those more fundamental thermodynamic properties. Better to transform this into a non-member function in `Thermodynamics/Utils.hpp`: 

~~~c++
auto phaseCompressibilityFactors(const ChemicalProperties&) -> ChemicalVector;
~~~

tests/regression/test_equilibrium_liq_gas.py
```diff
@@ -59,13 +59,13 @@ def test_equilibrium_CH4_liq_gas(temperature, pressure, num_regression):
     
     options = EquilibriumOptions()
     options.hessian = GibbsHessian.Exact
-    options.nonlinear.max_iterations = 100
-    options.optimum.max_iterations = 200
-    options.optimum.ipnewton.step = StepMode.Conservative
-    options.optimum.tolerance = 1e-17
+    options.optimum.max_iterations = 2000
```
allanleal, line 62: If the solver did not converge up to 200, then chances are that it won't for 2000! ;)

tests/regression/test_pedersen63.py
```diff
@@ -0,0 +1,379 @@
+import numpy as np
```
allanleal, line 1: I'm concerned with the length of this test and its future maintainability. It even has plots. This should perhaps be broken down into a tutorial/example script and a regression test. We should aim for simpler tests so that we don't have to spend a lot of time on it in the future if it breaks.

tests/regression/test_whitson18.py
```diff
@@ -0,0 +1,374 @@
+import numpy as np
```
allanleal, line 1: Same argument here about aiming for simpler and more straightforward tests.

allanleal (changes requested):
Thanks @volpatto and @alexandrebbruno . Please have a look at my comments and suggestions.

tests/regression/test_pedersen63.py
volpatto: Yes, indeed, the test is very long, I understand your point and I would like to share my opinion in some points.

The tutorial is a great idea, we can write one.

The idea and purpose of this test is to aim at "validation" and it is more like an integration test. The Reaktoro results are compared with results from well established literature. It has some "regressions", but regressions are not aimed to "validate", just to verify if a behavior is consistent through changes, even if the behavior is, well, wrong (when the regression test was inappropriately designed), i.e., regressions compare Reaktoro with itself. The idea of this test is different, it is to assert that Reaktoro result is valid against references.

In this sense, I think we could break this test in more tests, one for regressions, other for validation/verification.

The part of "plotting" can be removed, it is there just to check visually if a flag is set.

In my opinion, tests like this one is important and welcome, because it adds confidence to Reaktoro. But we can simplify and even remove the comparisons, turn it to simple regression tests as you suggested, no problem.

tests/regression/test_pedersen63.py
```diff
@@ -0,0 +1,379 @@
...
+
+    # Retrieve results by phase
+    df_pvtlib_fugacities_gas = df_pvtlib_fugacities[df_pvtlib_fugacities.Phase == 1]
+    df_pvtlib_fugacities_gas.reset_index(drop=True, inplace=True)
+    df_pvtlib_fugacities_liq = df_pvtlib_fugacities[df_pvtlib_fugacities.Phase == 2]
+    df_pvtlib_fugacities_liq.reset_index(drop=True, inplace=True)
+    df_pvtlib_composition_gas = df_pvtlib_compositions[df_pvtlib_compositions.Phase == 1]
+    df_pvtlib_composition_gas.reset_index(drop=True, inplace=True)
+    df_pvtlib_composition_liq = df_pvtlib_compositions[df_pvtlib_compositions.Phase == 2]
+    df_pvtlib_composition_liq.reset_index(drop=True, inplace=True)
```
volpatto, line 274: To simplify the test, I think that we could encapsulate parts like this one in fixtures.

tests/regression/test_pedersen63.py
```diff
@@ -0,0 +1,379 @@
...
+
+        plt.xlabel("Pressure [bar]")
+        plt.ylabel("Fugacities [bar]")
+        plt.title(f"Fixed T = {temperature} degC")
+        plt.legend(shadow=True)
+
+        plt.grid(True)
+
+        # plt.savefig("reaktoro_pvtlib_fugacities_liq_pedersen.png", dpi=300)
+        plt.show()
```
volpatto, line 379: I think that we could remove plots like this one, and put them in a tutorial. I think it would be good to create a ipynb and add it to `reaktoro-jupyter` repo. I could do it.

volpatto (commented):
Thanks for your review, @allanleal. I have added some comments to help our improvements/corrections.

tests/regression/test_pedersen63.py
allanleal: I think the fixture would make more sense if this becomes used all over the place, but I don't think it's the case. Let me know.

volpatto: Indeed, it is just a unique use. I was think to put it in a fixture to make the test at least clearer and smaller. This part will never break as soon as `pandas` does not break. It is just a "data gathering" step.

tests/regression/test_pedersen63.py
allanleal: Absolutely. We should have validation tests (like this one) as well (in addition to regression). 

The idea of shorter and more straightforward tests is to safeguard whoever is developing/contributing to Reaktoro (or its dependencies) and then needs to fix broken regression tests as a result of some code changes. But I understand sometimes simple enough is the way to go.

volpatto: Yeah, I agree.

I will see with the guys how we can improve the tests and simplify them.

tests/regression/test_pedersen63.py
allanleal: Better to keep these parts together then. We need to have tests structured/defined in a very intuitive way (so that everyone should be able to fix them quickly in case they break in the future). Moving parts to different locations will make them more difficult to understand. If a computed object needs to be used by many tests and its computation is time-consuming, then we move this to a single location (fixture).

Reaktoro/Core/ChemicalProperties.cpp
alexandrebbruno: Done,
move this function to Thermodynamics/ThermodynamicsUtils.hpp:

tests/regression/test_equilibrium_liq_gas.py
alexandrebbruno: yep! I changed back to 200 and the test converged

tests/regression/test_pedersen63.py
alexandrebbruno: Hey guys,

I think I've found a nice solution for both tests.
Solution:
- diego will make a demo demo for reaktoro-jupyter with the plot that I've removed from the test
- also tried to make those tests simpler and more straight foward

If you guys are still thinking that the test is complicated, I can try to remove the comparison with the pvtlib and only do the validation using regression

Reaktoro/Thermodynamics/ThermodynamicsUtils.cpp
```diff
@@ -0,0 +1,33 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+#include "ThermodynamicsUtils.hpp"
+
+#include <Reaktoro/Common/Constants.hpp>
+
+namespace Reaktoro {
+
+auto phaseCompressibilityFactors(const ChemicalProperties& properties) -> ChemicalVector
```
allanleal, line 24: This file should be `Thermodynamics/Utils.hpp/cpp`. Repeated file names are OK, as long as they are in different directories. I'm now wondering however why we need this function. It is a simple computation and also specific to systems/EOS formulations that require compressibility factors -- not something generally applicable. It's also not used in any other places. Shall we remove this function until we are clear we really need it in the library?

allanleal:
Hi @alexandrebbruno - many thanks for revising this PR. All is fine now. Please see however my point about not considering the `phaseCompressibilityFactors` function in this PR.

Reaktoro/Thermodynamics/ThermodynamicsUtils.cpp
volpatto: > It is a simple computation and also specific to systems/EOS formulations that require compressibility factors -- not something generally applicable.

Agreed. We can remove it, as we can easily calculate post-processing `Reaktoro`'s results. The `Z`-factor is important in PVT analysis and hydrocarbon systems, but it is simple to compute it with `Reaktoro::ChemicalProperties`. IMHO, we can remove it for now.

volpatto: Ah, one more thing: in general, the Z-factor is used to validate Cubic EoS results, since it is a key result for such EoSs. I don't remember if I added a test comparing Z results, but I did this study.

tests/regression/test_pedersen63.py
volpatto: Great, @alexandrebbruno! Thanks!

tests/regression/test_equilibrium_liq_gas_aq/test_equilibrium_CH4_H2S_CO2_H2O_liq_gas_aq_temperature_equal_310_95_K_and_62_6_bar_.csv
```diff
@@ -1,2 +1,2 @@
 ,CO2(aq),H2S(aq),H2O(l),CH4(g),CO2(g),H2S(g),H2O(g),CH4(liq),CO2(liq),H2S(liq),H2O(liq)
-0,0.0026002573995516864,0.011477541939109524,0.48184939865204945,0.0082682414544814273,0.0026002573995516864,0.010202123420894117,4.7543713579791212e-05,0.041731758545518569,0.046277437221844242,0.37832033463999637,0.018103057634370923
+0,9.897014523419688e-23,0.011361212608224843,0.48236568923000694,1.0047813446754076e-22,9.897014523419688e-23,1.001374498416919e-22,9.4702520328251751e-23,0.050000000000000024,0.048877250212860657,0.38863878739177515,0.017634310769993146
```
volpatto, line 2: @alexandrebbruno do you still have the code that generates phase envelopes? If it is easy, can you generate one for this case? Don't worry, it can be after this PR.

Reaktoro/Thermodynamics/ThermodynamicsUtils.cpp
alexandrebbruno: ok, it doesn't seem to be de right time to add this function. I will remove it.

- [ ] remove `phaseCompressibilityFactors`

alexandrebbruno: done

tests/regression/test_equilibrium_liq_gas_aq/test_equilibrium_CH4_H2S_CO2_H2O_liq_gas_aq_temperature_equal_310_95_K_and_62_6_bar_.csv
alexandrebbruno: I don't have it now :/
But can try to compute the phase envelope of this case as soon as I have some free time, probably the next weekend

allanleal (approved):

PR #204, opened 2021-07-14
First cpp and python examples

mtsveta:
First cpp and python examples to demonstrate Reaktoro v2.0 functionality.


examples/cpp/ex-equilibrium-co2-solubility-nacl-h2o-fixed-ph.cpp
```diff
@@ -0,0 +1,89 @@
...
+    conditions.startWith("H2O(aq)", 1.0, "kg");
+    conditions.startWith("Na+",     1.0, "mol");
+    conditions.startWith("Cl-",     1.0, "mol");
+    conditions.startWith("CO2(g)", 10.0, "mol");
+
+    // Define initial equilibrium state
+    ChemicalState state(system);
+
+    // Equilibrate given initial state
+    solver.solve(state, conditions);
```
allanleal, line 69: We can write below:

~~~c++
std::cout << state << std::endl;
~~~

and remove all the rest. 

Also, `ChemicalState` is not very revealing: only temperature, pressure, species amounts. We can also create an object of `ChemicalProps` and another of `AqueousProps` and print them:

~~~c++
// Compute chemical and thermodynamic properties at equilibrium state
ChemicalProps props(state);

std::cout << props << std::endl;

// Compute chemical properties for the aqueous phase at equilibrium state
AqueousProps aqprops(props);

std::cout << aqprops << std::endl;
~~~

Something like this.

examples/cpp/ex-equilibrium-co2-solubility-nacl-h2o-fixed-ph.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+int main()
+{
+    // Initialize a thermodynamic database
+    Database db("supcrtbl.yaml");
+
+    // Define aqueous phase with selected species
+    AqueousPhase aqueousphase("H2O(aq) CO2(aq) CO3-2 Cl- H+ H2(aq) HCO3- Na+ NaCl(aq) NaOH(aq) O2(aq) OH- HCl(aq)");
+    aqueousphase.setActivityModel(chain(
+            ActivityModelHKF(),
```
allanleal, line 29: Ensure 4-space indentation. You may want to configure your editor for this.

examples/cpp/ex-equilibrium-custom-database.cpp
```diff
@@ -23,24 +23,33 @@ auto createCustomDatabase() -> Database;
...
     state.setTemperature(25.0, "celsius");
     state.setPressure(1.0, "bar");
     state.setSpeciesMass("H2O", 1.0, "kg");
 
+    // Define equilibrium solver and equilibrate given initial state
     EquilibriumSolver solver(system);
-
     solver.solve(state);
 
+    // Obtain species composition from the equilibrated state
```
allanleal, line 46: Here too, we print the `ChemicalState` using `operator<<`. We should have a method `ChemicalState::output` like in the v1.x, which outputs to a file.

~~~c++
state.output("state.txt");
~~~

because printing to the terminal can be inconvenient to some.

examples/cpp/ex-equilibrium-supcrt-database.cpp
```diff
@@ -20,22 +20,29 @@ using namespace Reaktoro;
 
 int main()
 {
+    // Initialize a thermodynamic database
     SupcrtDatabase db("supcrt98.xml");
```
allanleal, line 24: The `SupcrtDatabase` will disappear, as we no longer need it because of the new `Database` class. We should instead use `Database db("supcrtbl.yaml");` here, which is a more recent version of SUPCRT (SUPCRTBL) and using YAML format (instead of XML in `SupcrtDatabase`).

allanleal (commented):
It's going very well. Thanks a lot for this. See my comments above.

examples/cpp/ex-equilibrium-co2-solubility-nacl-h2o-fixed-ph.cpp
mtsveta: Corrected

examples/cpp/ex-equilibrium-co2-solubility-nacl-h2o-fixed-ph.cpp
mtsveta: Replaced

examples/cpp/ex-equilibrium-supcrt-database.cpp
mtsveta: Deleted one cpp example and replaced  one python example with different equilibirum test-case using yaml database

examples/cpp/ex-equilibrium-custom-database.cpp
mtsveta: Implemented method `state.output("state.txt")` in the `ChemicalState` class and replaced it the majority of the cpp examples

examples/python/ex-equilibrium-carbonates-solubility-seawater.py
```diff
@@ -0,0 +1,109 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+# Import reaktoro python package
+from reaktoro import *
+import numpy as np
+
+# Function to calculate equilibrium of the carbonates and seawater
+def carbonates_in_seawater(system, T, P):
```
allanleal, line 23: This function should be defined after ChemicalState and EquilibriumSolver are created to avoid overhead. In Debug build, this example is rather slow.

examples/python/ex-equilibrium-carbonates-solubility-seawater.py
```diff
@@ -0,0 +1,109 @@
+# Reaktoro is a unified framework for modeling chemically reactive systems.
+#
+# Copyright (C) 2014-2020
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
```
allanleal, line 8: Note I'm adding this Acknowledgements section here in the examples. Feel free to alter to a better format.

examples/cpp/ex-equilibrium-calcite-solubility-rainwater.cpp
```diff
@@ -15,6 +15,16 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this library. If not, see <http://www.gnu.org/licenses/>.
 
+// -----------------------------------------------------------------------------
```
allanleal, line 18: Note I'm adding this Acknowledgements section here in the examples. Feel free to alter to a better format.

allanleal: You'll need to go inside the file. It's not showing here in the Conversation tab of GitHub.

examples/python/ex-equilibrium-carbonates-solubility-seawater.py
allanleal: You'll need to go inside the file. It's not showing here in the Conversation tab of GitHub.

allanleal:
Svetlana, very good! The only critical issue that remains (but it should be easy to fix) is to alter those examples in which a function is  defined to perform a calculation and inside this function an `EquilibriumSolver` object is created every evaluation. This is expensive, as there is some overhead in the creation of this solver. Define the function after the creation of the `EquilibriumSolver` (or a class that has this EquilibriumSolver as a data member). The easiest is to move the function definition after EquilibriumSolver is created and then the function uses it (the function scope can access this object from outside).

allanleal:
Just one more thing. :)
The examples are using `supcrt98.yaml`. Could you please verify if using `supcrtbl.yaml` works (produce similar results) and if so we use `supcrtbl.yaml` instead?

allanleal:
I'll merge this now, and the tasks above done later. :+1:

PR #205, opened 2021-07-14
Move molar_surface_area to inside a lambda function

volpatto:
This is a necessary change in order to propagate the changes to `Reaktoro::KineticSolver` through MineralReaction in Reactive Transport problems considering kinetics and updating -- for instance, due to dissolution -- the specific surface area (accessed and modified through a `Reaktoro::MineralReaction` `std::shared_ptr`).

CC: @tadeu

### DIFF
--- a/Reaktoro/Thermodynamics/Reactions/MineralReaction.cpp
+++ b/Reaktoro/Thermodynamics/Reactions/MineralReaction.cpp
@@ -489,14 +489,11 @@ auto createReaction(const MineralReaction& mineralrxn, const ChemicalSystem& sys
     }
     else
     {
-        // The molar surface area of the mineral
-        const double molar_surface_area = molarSurfaceArea(mineralrxn, system);
-
-        // The surface area of the mineral
-        const double surface_area = mineralrxn.surfaceArea();
-
         rate = [=](const ChemicalProperties& properties) mutable
         {
+            // The molar surface area of the mineral
+            const double molar_surface_area = molarSurfaceArea(mineralrxn, system);
+
             // The composition of the chemical system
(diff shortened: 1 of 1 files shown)


Reaktoro/Thermodynamics/Reactions/MineralReaction.cpp
```diff
@@ -489,14 +489,11 @@ auto createReaction(const MineralReaction& mineralrxn, const ChemicalSystem& sys
     }
     else
     {
-        // The molar surface area of the mineral
-        const double molar_surface_area = molarSurfaceArea(mineralrxn, system);
-
-        // The surface area of the mineral
-        const double surface_area = mineralrxn.surfaceArea();
```
volpatto, line 496: I have removed this one since it is not used in the code that follows.

Reaktoro/Thermodynamics/Reactions/MineralReaction.cpp
```diff
@@ -489,14 +489,11 @@ auto createReaction(const MineralReaction& mineralrxn, const ChemicalSystem& sys
...
-
         rate = [=](const ChemicalProperties& properties) mutable
         {
+            // The molar surface area of the mineral
+            const double molar_surface_area = molarSurfaceArea(mineralrxn, system);
+
             // The composition of the chemical system
             const auto n = properties.composition();
```
allanleal, line 499: OK, I get it. This evaluation should be inside the lambda so that when you alter `mineralrxn` externally this change will 
cause the `molar_surface_area` to change as well.

allanleal (approved):

PR #209, opened 2021-08-06
Fix KineticSolver::setPartition when Inert species are set

hfcpeixoto:
When Inert species are set on the partition, `KineticSolver::setPartition()` breaks due to a wrong number of columns on the coefficients' matrix `B`, which is implemented to consider all species on the system (`ChemicalSystem::numSpecies()`).

This PR aims to fix this behaviour, by taking into account that the system may have Inert species, which should be disregarded on the coefficients' matrix number of columns. The number of columns, thus, must be the sum of Equilibrium and Kinetic species in the system.

### DIFF
--- a/tests/regression/test_kinetic_solver.py
+++ b/tests/regression/test_kinetic_solver.py
@@ -9,7 +9,9 @@ from reaktoro import (
     Database,
     equilibrate,
     EquilibriumProblem,
+    EquilibriumSolver,
     KineticPath,
+    KineticSolver,
     Partition,
     ReactionSystem,
 )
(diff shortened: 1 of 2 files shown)


tests/regression/test_kinetic_solver.py
```diff
@@ -314,3 +316,76 @@ def test_kinetic_path_solve_final_state(
     path.solve(state, time_span.ti, time_span.tf, time_span.unit)
 
     state_regression.check(state, default_tol=dict(atol=1e-5, rtol=1e-14))
+
+
+@pytest.mark.parametrize(
+    "setup, time_span, minerals_to_add",
```
tadeu, line 322: Perhaps only `time_span` parameter is necessary here, since the other parameters don't seem to vary?

hfcpeixoto: True that. I was using other method,

tadeu (approved):

allanleal (approved):

allanleal:
Thanks, @hfcpeixoto for the fix and @tadeu for the review!

PR #213, opened 2021-09-11
Ion exchange implementation

mtsveta:
This PR implements a mechanism of ion exchange on the surface of the exchanger species. The exchanger and ion exchange phases are formed by the species collected from the PHREEQC database. Creation of the phases replies on the functionality provided by the class `GenericPhase`.

The current implementation assumes only one exchanger (X-) and ion exchange phase formed by respective species (NaX, KX, CaX2, MgX2, etc).

For the activity model of the ion exchange, the Gaines-Thomas activity model is used. 

Tasks:

- [x] Implement `ActivityModelIonExchange()` function.
- [x] Implement ion exchange cpp example.
- [x] Implement cpp tests in `ActivityModelIonExchange.test.cxx`
- [x] Implement python binding for the `ActivityModelIonExchange` in `ActivityModelIonExchange.py.cxx` file
- [x] Implement py tests in `ActivityModelIonExchange.py`

### DIFF
--- a/Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
+++ b/Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
@@ -30,9 +30,15 @@ namespace detail {
 // Return the number of exchanger's equivalents (the charge of cations) in the ion exchange species.
 auto exchangerEquivalentsNumber(const Species& species) -> real
 {
+    // Run through the elements of the current species and return the coefficient of the exchanger
     for(auto [element, coeff] : species.elements())
         if(!Elements::withSymbol(element.symbol()))
             return coeff;
+    // If all the elements are part of the periodic table then the exchanger is missing
+    warning(1, "Could not get information about the exchanger equivalents number. "
+             "Ensure the ion exchange phase contains correct species");
+    return 0;
+
 }
 /// Return the InoExchangeActivityModel object based on the Gaines--Thomas model.
 auto activityModelIonExchangeGainesThomas(const SpeciesList& species) -> ActivityModel
@@ -55,6 +61,9 @@ auto activityModelIonExchangeGainesThomas(const SpeciesList& species) -> Activit
         // Auxiliary references
         auto& ln_g = props.ln_g;
         auto& ln_a = props.ln_a;
+
+        // Export the aqueous mixture and its state via the `extra` data member
+        props.extra = { ze };
 
         // Calculate the ln of equivalence fractions
         ArrayXr ln_beta = (x * ze / (x * ze).sum()).log();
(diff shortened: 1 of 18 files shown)


Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
```diff
@@ -30,9 +30,15 @@ namespace detail {
 // Return the number of exchanger's equivalents (the charge of cations) in the ion exchange species.
 auto exchangerEquivalentsNumber(const Species& species) -> real
 {
+    // Run through the elements of the current species and return the coefficient of the exchanger
     for(auto [element, coeff] : species.elements())
         if(!Elements::withSymbol(element.symbol()))
             return coeff;
+    // If all the elements are part of the periodic table then the exchanger is missing
+    warning(1, "Could not get information about the exchanger equivalents number. "
```
allanleal, line 38: I think here we should then raise an error instead. Please use `errorif` and `warningif` variants instead of `error` and `warning` (they are more efficient in handling the message (only actually processed if the condition is `true`), and these ones will be gone at some point).

Also, please use `true` instead of `1`.

Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
```diff
@@ -0,0 +1,94 @@
...
+        auto& ln_a = props.ln_a;
+
+        // Export the aqueous mixture and its state via the `extra` data member
+        props.extra = { ze };
+
+        // Calculate the ln of equivalence fractions
+        ArrayXr ln_beta = (x * ze / (x * ze).sum()).log();
+
+        // Calculate the ln of activity coefficients
+        ln_g = ArrayXr::Zero(num_species);
```
allanleal, line 70: This is currently zero, but I assume we will fill out some activity coefficients in the future, right (e.g., based on Davies or Debye-Huckel)

examples/cpp/ex-equilibrium-phreeqc-ion-exchange.cpp
```diff
@@ -0,0 +1,110 @@
...
+    PhreeqcDatabase db("phreeqc.dat");
+
+    // Define aqueous phase
+    AqueousPhase aqueousphase(speciate("H O C Ca Na Mg Cl"));
+    aqueousphase.setActivityModel(chain(
+            ActivityModelHKF(),
+            ActivityModelDrummond("CO2")
+            ));
+
+    // Fetch species for ion-exchange modeling
```
allanleal, line 54: We should have a PR in the future that simplifies this process. We would need IonExchangerPhase and IonExchangePhase. With this, we would automate the collection of relevant species.

Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
mtsveta: Yes, exctaly!

examples/cpp/ex-equilibrium-phreeqc-ion-exchange.cpp
mtsveta: Completely agree! I have already started with `IonExchangePhase` actually, I will include this into another PR.

PR #217, opened 2021-09-27
Fix segmentation fault error in aqueous properties calculation and output

mtsveta:
Fix segmentation fault error generated during the call

~~~c++
AqueousProps aqprops(props);
td::cout << aqprops << std::endl;
~~~

on the stage of pE calculation and during the output.

Tasks:

- [x] Fix segmentation fault error during the call `pE()`.
- [x] Fix segmentation fault error during the call of `<<operator()`.

New Tasks: 
- [x] Implement tests for `assembleFormulaMatrix`, from `Core/Utils.hpp`, in `Core/Utils.test.cxx`
- [x] Implement tests for `AqueousProps` in `AqueousProps.tests.cxx`

**Note:** 
- For the `assembleFormulaMatrix` test, check `ChemicalSystem.test.cxx`, the end of the file.
- For the `AqueousProps` test, also check `ChemicalSystem.test.cxx` to see the line `ChemicalSystem system = test::createChemicalSystem();` for a mock `ChemicalSystem` object.

### DIFF
--- a/Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
+++ b/Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
@@ -191,7 +191,7 @@ struct AqueousProps::Impl
 
     auto elementMolalities() const -> VectorXr
     {
-        const auto E = system.phase(indexAqueousPhase(system)).elements().size();
+        const auto E = system.elements().size();
         const auto& m = aqstate.m.matrix();
         return Aaq.topRows(E) * m;
     }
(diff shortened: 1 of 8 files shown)


Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
```diff
@@ -191,7 +191,7 @@ struct AqueousProps::Impl
 
     auto elementMolalities() const -> VectorXr
     {
-        const auto E = system.elements().size();
+        const auto E = system.phase(0).elements().size();
```
allanleal, line 194: This presumes the aqueous phase is always the first phase in the system, which may not be the case.

allanleal (changes requested):

Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
mtsveta: Okay, I thought that it might be not too general. I have corrected this line.

allanleal:
Awesome, Svetlana!

PR #219, opened 2021-09-30
Little changes in creation and calculation of the ion exchange 

mtsveta:
This PR extends tests and example lists related to the ion exchange modeling. 

Tasks:

- [x] Extend the list of constructors to create `IonExchangePhase`.
- [x] Add python binding for `IonExchangePhase`.
- [x] Add corresponding cpp- and python-tests for `IonExchangePhase`.
- [x] Add python examples illustrating working with `phreeqc.dat` and model ion exchange with it.
- [x] Change method of calculating exchanger equivalencies in `IonExchangeSurface` class, making it independent on `Elements` singleton.
- [x] Extend cpp-tets with cases based on the custom database.


Reaktoro/Core/Phases.test.cxx
```diff
@@ -634,4 +665,68 @@ TEST_CASE("Testing Phases", "[Phases]")
...
         checkMineralPhase(phasevec[3], "Graphite");
     }
+
+    //=================================================================================================================
+    //-----------------------------------------------------------------------------------------------------------------
+    // TESTING CLASS: IonExchangePhase
+    //-----------------------------------------------------------------------------------------------------------------
+    //=================================================================================================================
+
+    auto speciesListToStringList = [&](const SpeciesList& specieslist) -> StringList
```
allanleal, line 675: Check later `vectorize` in Algorithms.hpp and RKT_LAMBDA

allanleal: This method is also being used in many replaces and repeated implementation. We could have a method in `Core/Utils.hpp` called `extractNames(const SpeciesList&)` that does this. Then also export this in `Core.py.cxx` so that it can be used in the python scripts. Please also implement overload versions for `ElementList` and `PhaseList`.

Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.cpp
```diff
@@ -112,6 +111,7 @@ auto activityModelIonExchangeGainesThomas(const SpeciesList& species) -> Activit
         }
         // Add the correction introduced by the activity coefficients
         ln_a += ln_g;
+
```
allanleal, line 114: Empty line

Reaktoro/Thermodynamics/Surface/ActivityModelIonExchange.test.cxx
```diff
@@ -57,69 +57,62 @@ inline auto initializeAqueousMoleFractions(const SpeciesList& species) -> ArrayX
     return n / n.sum();
 }
 
+namespace test
```
allanleal, line 60: `namespace test {} // namespace test`

Also, no indentation inside the namespace.

Reaktoro/Thermodynamics/Surface/IonExchangeSurface.cpp
```diff
@@ -60,16 +59,16 @@ struct IonExchangeSurface::Impl
     }
 
     // Return the number of exchanger's equivalents (the charge of cations) for the ion exchange species.
-    static auto exchangerEquivalentsNumber(const Species& species) -> real
+    static auto exchangerEquivalentsNumber(const Species& species, const String& exchanger_symbol) -> real
```
allanleal, line 62: Why is this method `static`? Should we move this outside the class?

Reaktoro/Thermodynamics/Surface/IonExchangeSurface.cpp
```diff
@@ -82,9 +81,12 @@ struct IonExchangeSurface::Impl
         // The numbers of exchanger's equivalents for exchange species
         ze = ArrayXr::Zero(num_species);
 
+        // Define the element symbol presenting the exchanger
+        auto exchanger_symbol = species[this->idx_exchanger].elements().symbols()[0];
```
allanleal, line 85: There seems to be an implicit convention here that the first element is the exchanger. I think this should be implemented in a generic way, to avoid possible issues/strange behavior later that will be difficult to spot.

allanleal, line 85: No need for `this->` here.

Reaktoro/Core/Phases.test.cxx
mtsveta: Done.

Reaktoro/Thermodynamics/Surface/IonExchangeSurface.cpp
mtsveta: I have removed `static`, it is not needed here.

Reaktoro/Thermodynamics/Surface/IonExchangeSurface.cpp
mtsveta: Corrected.

allanleal:
Super, thanks, Svetlana!

PR #233, opened 2022-01-25
Initial state of the EquilibriumSolver

mtsveta:


mtsveta:
This PR is no longer up do date

PR #235, opened 2022-01-31
Add missing ln(xw) term in Debye-Huckel model 

allanleal:
This PR adds a missing `ln(xw)` term in the activity coefficients computed with the Debye-Huckel model.

Thanks to @volpatto for pointing out to this!


allanleal:
Closing this for now.

PR #245, opened 2022-04-29
Fb steps 146 fallback long range only

caiofcm:
Too features were implemented:

- Only calculating short-range contribution for species with e-uniquac parameters available
- Options to select the long-range model. 
  - Original Thomsen DH
  - DH model using PHREEQC parameters (using the existing model in reaktoro)

Because the existing model was used, few adjustments were needed. For instance, e-uniquac operates using mol fraction scale. Only at the end the scale is changed to molality. Therefore, I changed the scale of DH using phreeqc to mol scale. Such that at the end of the e-uniquac model it will be added to the short-range contribution (in mol scale) and later will be converted to molalility scale.

There are some questions in the comments to be discussed.


caiofcm:
Wrong target for the PR! Thus, I am closing it.

PR #246, opened 2022-05-07
Four Python examples did not run; made them run in PyCharm IDE

cskv:


allanleal:
Hi @cskv , thanks a lot for the fixes!

PR #250, opened 2022-05-26
Added checks on electroneutrality and conservation of mass when constructing a reaction

cardinalgeo:


Reaktoro/Core/Utils.cpp
```diff
@@ -195,5 +195,15 @@ auto extractNames(const PhaseList& list) -> Strings
...
 }
 
+auto isChargeBalanced(const Reaction& reaction) -> bool
+{
+    return true; // replace by code
+}
+
+auto isElementBalanced(const Reaction& reaction) -> bool
+{
+    return true; // replace by code
```
allanleal, line 205: Please do the same here.

Reaktoro/Core/Database.cpp
```diff
@@ -30,6 +30,7 @@ CMRC_DECLARE(ReaktoroDatabases);
 #include <Reaktoro/Common/Exception.hpp>
 #include <Reaktoro/Common/ParseUtils.hpp>
 #include <Reaktoro/Core/Support/DatabaseParserYAML.hpp>
+#include <Reaktoro/Core/Utils.hpp>
```
allanleal, line 33: With this include, we are able to import those we just declared: `isChargeBalanced` and `isElementBalanced`.

Reaktoro/Core/Utils.test.cxx
```diff
@@ -147,4 +148,19 @@ TEST_CASE("Testing CoreUtils", "[CoreUtils]")
             CHECK(A_min.row(i) == A_min_expected.row(i));
     }
 
+    SECTION("Testing isChargeBalanced and isElementBalanced for reactions")
+    {
+        auto reaction1 = Reaction().withEquation("H2O = H+ + OH-");
```
allanleal, line 153: Please add more reactions and tests below as needed.

Reaktoro/Core/Database.cpp
```diff
@@ -113,7 +114,13 @@ struct Database::Impl
         Pairs<Species, double> reactants;
         for(const auto [name, coeff] : pairs)
             reactants.emplace_back(species.get(name), coeff);
-        return Reaction().withEquation(reactants);
+
+        auto reaction = Reaction().withEquation(reactants);
```
allanleal, line 118: We create the `Reaction` object here and veryfy its correctness on the lines below. Using `warningif` instead of `errorif`.

Reaktoro/Core/Utils.hpp
```diff
@@ -28,6 +28,7 @@ class ChemicalSystem;
 class ElementList;
 class Phase;
 class PhaseList;
+class Reaction;
```
allanleal, line 31: This declaration is needed so that we can refer to the `Reaction` symbol in this file. We could have instead included the file `Reaktoro/Core/Reaction.hpp`, but forward declarations are a C++ trick to help with less compilation work when a header file changes. For example, if we included `Reaktoro/Core/Reaction.hpp` here, and this `Utils.hpp` file is included in many other cpp files, all these files would have to be compiled. This would not happen if instead we use this forward declaration trick.

PR #254, opened 2022-07-17
WIP: Surface complexation functionality

mtsveta:
This PR implements a mechanism of surface complexation on the charged surface. The surface complexation phases are formed by the species collected from the PHREEQC database. The creation of the phases replies to the functionality provided by the class `GenericPhase`.

The current implementation assumes only the default surface `Hfo` with two sites `Hfo_s` and `Hfo_w` and the species sorbed on them (defined by the PHREEQC databases).

For the activity model of the surface complexation, the no double layer and double layer base on Dzombak and Morel 1990 are implemented.

Tasks:

- [x] Extend class `Phasses` with `DoubleLayerPhase` and `SurfaceComplexationPhase` classes.
- [x] Implement `ComplexationSurface`,`ComplexationSurfaceSite`, and `DoubleLayer` classes to be used for the activity models used in `DoubleLayerPhase` and `SurfaceComplexationPhase`.
- [x] Implement `ActivityModelSurfaceComplexation` and `ComplexationSurface` classes with activity models described above.
- [x] Implement `DoubleLayerProps` and `ComplexationSurfaceProps` classes to output the properties of the double layer and complexion surface, respectively.
- [x] Implement surface complexation cpp and python tests.
- [x] Implement python bindings for implemented above functionality.
- [x] Implement surface complexation cpp and python examples.
- [x] Add cosmetic and type corrections to other classes.


Reaktoro/Thermodynamics/Aqueous/AqueousMixture.hpp
```diff
@@ -51,6 +51,10 @@ struct AqueousMixtureState
 
     /// The stoichiometric molalities of the ionic species (in mol/kg)
     ArrayXr ms;
+
+    /// The charges of the aqueous species
+    ArrayXr z;
```
allanleal, line 56: The charges of the species should not be here because they are constant, from state to state. They are available in the `AqueousMixture` class, `AqueousMixture::charges`.

Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
```diff
@@ -49,14 +49,16 @@ namespace {
...
     const auto aqueous_phases = system.phases().withAggregateState(AggregateState::Aqueous);
-    warning(aqueous_phases.size() > 1,
-        "While creating an AqueousProps object, it has been detected ",
-        "more than one aqueous phase in the system. The AqueousProps object "
-        "created will correspond to the first aqueous phase found.");
+//    warning(aqueous_phases.size() > 1,
+//        "While creating an AqueousProps object, it has been detected ",
+//        "more than one aqueous phase in the system. The AqueousProps object "
+//        "created will correspond to the first aqueous phase found.");
     const auto idx = system.phases().findWithAggregateState(AggregateState::Aqueous);
```
allanleal, line 56: Let's talk about this here. The original convention was to pick the first phase with `AggregateState::Aqueous`. I think this is more generic than by finding with name `"AqueousPhase"`, which the user may change.

Reaktoro/Thermodynamics/Aqueous/AqueousProps.hpp
```diff
@@ -89,6 +89,9 @@ class AqueousProps
     /// Return the molality concentrations of the species (in molal).
     auto speciesMolalities() const -> ArrayXr;
 
+    /// Return the charges of the species.
+    auto speciesCharges() const -> ArrayXr;
```
allanleal, line 93: The electric charges of the species could become private member data in `AqueousProps` and we return a const-reference to it, `ArrayXrConstRef`.

Reaktoro/Thermodynamics/Aqueous/AqueousMixture.hpp
mtsveta: Charges of species are removed from `AqueousMixtureState`, `ComplexationSurfaceState` that effected calculation of the `AqueousProps`,   `DoubleLayerProps`, and `ComplexationSurfaceProps`. See commit 94b8c2ed717c07ae8403c3a317d89ece969aae92.

Reaktoro/Thermodynamics/Aqueous/AqueousProps.cpp
mtsveta: Agree. However, is it general enough for the case when there are more then one aqueous phases. Or is this scenario is unlikely?

Reaktoro/Thermodynamics/Aqueous/AqueousProps.hpp
mtsveta: Done in commit fbe98c0cee732e537bf5463a064adea0f20bc1f9.

Reaktoro/Equilibrium/EquilibriumSetup.cpp
```diff
@@ -140,13 +140,22 @@ struct EquilibriumSetup::Impl
     /// Assemble the coefficient matrix `Aex` in optimization problem.
     auto assembleMatrixAex() const -> MatrixXd
     {
+//        // Amount of species in aqueous phase (assuming that it is the first defined phase)
```
allanleal, line 143: We can remove these commented lines, because in the kinetics branch I'm working on these additional linear constraints can be easily imposed via EquilibriumSpecs.

mtsveta: Done in the commit e64e0a54e950ef4b36450c00ec410edbf5cf3b56

PR #303, opened 2023-06-14
Fix wrong check for K (Kelvin) in a pressure function

hfcpeixoto:
Function `EquilibriumProblem::startWithPressure(real value, Chars unit)` tries to convert `unit` to `"K"` (Kelvin), which is a temperature unit, not a pressure one.

As pressures are handled in `"Pa"` in other methods in `EquilibriumProblem`, conversion to `"Pa"` is proposed as a fix.


hfcpeixoto:
Yeah, `EquilibriumSolver` seems to cope better with `EquilibriumProblem`'s base classes `EquilibriumConditions` and `EquilibriumRestrictions`. 

However, as the code I am working on is moving from Reaktoro v1 to v2, and it used quite a lot of features from `EquilibriumProblem`, using the updated version of this class seems to be the way to go. 

Should we keep the PR?

PR #339, opened 2023-09-18
Emit warning and/or error when ionic strength is high for `ActivityModelPhreeqc`

allanleal:
This pull request implements a warning if `ActivityModelPhreeqc` is used with an aqueous solution with ionic strength above 6 molal, suggesting the use of `ActivityModelPitzer` instead but also recognizing that the high ionic strength could be a result of an intermediate state during the equilibrium calculation.

Additionally, this pull request implements an error check that will throw a runtime error if the water mole fraction is below 0.48549234635, which causes the water activity to be negative according to the equation used by PHREEQC:

image

**Reference:** Parkhurst, DL, & Appelo, CAJ (1999). User's Guide for PHREEQC (Version 2) --- A computer program for speciation, batch reaction, one-dimensional transport, and inverse geochemical calculations. USGS Water Resources Investigations Report, 99(4259), 326.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include(CCache)
 include(CMakeRC)
 
 # Set the name of the project, its version and other information
-project(Reaktoro VERSION 2.9.1 LANGUAGES CXX)
+project(Reaktoro VERSION 2.9.2 LANGUAGES CXX)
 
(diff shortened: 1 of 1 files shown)


Jarnovr (approved):
Looks good to me!

This will be a useful message for our users.

PR #345, opened 2023-11-10
Implement alkalinity calculation in AqueousProps

volpatto:
## Overview

This PR implements the Alkalinity calculation in `AqueousProps` according to [Wolf-Gladrow et al. (2007)](https://doi.org/10.1016/j.marchem.2007.01.006).

## Note

Tests comparing against expected results from literature and exact values are provided to verify the implementation.

### DIFF
--- a/Reaktoro/Utils/AqueousProps.cpp
+++ b/Reaktoro/Utils/AqueousProps.cpp
@@ -32,9 +32,7 @@ using namespace tabulate;
 #include <Reaktoro/Common/Constants.hpp>
 #include <Reaktoro/Common/Enumerate.hpp>
 #include <Reaktoro/Common/Exception.hpp>
-#include <Reaktoro/Common/Warnings.hpp>
 #include <Reaktoro/Core/ChemicalProps.hpp>
-#include <Reaktoro/Core/ChemicalPropsPhase.hpp>
 #include <Reaktoro/Core/ChemicalState.hpp>
 #include <Reaktoro/Core/ChemicalSystem.hpp>
 #include <Reaktoro/Core/Phase.hpp>
--- a/Reaktoro/Utils/AqueousProps.cpp
+++ b/Reaktoro/Utils/AqueousProps.cpp
@@ -386,7 +386,7 @@ struct AqueousProps::Impl
         return res;
     }
 
-        auto alkalinity() const -> real
+    auto alkalinity() const -> real
     {
         // The number of species in the system
         const Index num_species = system.species().size();
--- a/Reaktoro/Utils/AqueousProps.cpp
+++ b/Reaktoro/Utils/AqueousProps.cpp
@@ -426,7 +426,8 @@ struct AqueousProps::Impl
         for(const auto& [alkalinity_factor, alkalinity_species_name] : factors_and_species)
         {
             // Get the index of the current ion
-            const Index i = system.species().findWithFormula(alkalinity_species_name);
+            const auto& aq_species_list = system.species().withAggregateState(AggregateState::Aqueous);
+            const Index i = aq_species_list.findWithFormula(alkalinity_species_name);
 
             // Calculate contribution to alkalinity if the species is in the system
             if (i < num_species) {
--- a/Reaktoro/Utils/AqueousProps.cpp
+++ b/Reaktoro/Utils/AqueousProps.cpp
@@ -181,8 +181,11 @@ struct AqueousProps::Impl
     /// The echelon form of the formula matrix `Aaqs` of the aqueous species.
     Optima::Echelonizer echelonizer;
 
-    // The chemical potential models for the non-aqueous species (as if they were pure phases) for the computation of their saturation indices.
+    /// The chemical potential models for the non-aqueous species (as if they were pure phases) for the computation of their saturation indices.
     Vec<Fn<real(ChemicalProps const&)>> chemical_potential_models;
+
+    /// The alkalinity contribution factors of some aqueous species based on the alkalinity model of Wolf-Gladrow et al. (2007)
+    Pairs<double, Index> alkalinity_factors;
 
     Impl(ChemicalSystem const& system)
     : system(system),
@@ -203,6 +206,9 @@ struct AqueousProps::Impl
 
         // The symbols of the elements in the aqueous phase
         const auto symbols = vectorize(phase.elements(), RKT_LAMBDA(x, x.symbol()));
+
+        // The aqueous species in the aqueous phase
+        auto const& aqspecies = phase.species();
 
         // Collect the species from the database that contains the elements in the aqueous phase
         const auto species_same_elements = system.database().species().withElements(symbols);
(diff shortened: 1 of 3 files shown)


volpatto:
@allanleal I have opened this PR here with improvements compared to the private one that you have already checked. Please feel free to share your thoughts and suggestions!

Reaktoro/Utils/AqueousProps.cpp
```diff
@@ -386,10 +386,63 @@ struct AqueousProps::Impl
         return res;
     }
 
-    auto alkalinity() const -> real
+        auto alkalinity() const -> real
```
volpatto:
```suggestion
  auto alkalinity() const -> real
```

Fix weird format.

volpatto: Fixed.

allanleal:
Hi Diego, thanks for this PR. We have some past work done on electrical conductivity and alkalinity that needs to be revived and reviewed, and in the process check how your contribution fits together. There are many definitions of alkalinity and equations proposed in the literature. We were trying to organize this into different models that the user could choose from (with parameter values, in the case of electrical conductivity, for example, provided via local or embedded param files). Also, the idea would be to produce alkalinity values using the same approach in PHREEQC (which would be the default model if not explicitly changed by the user), but it could be different (based on the definitions of Wolf-Gladrow et al.).

I suggest we keep this PR open for now.

volpatto:
> Hi Diego, thanks for this PR. We have some past work done on electrical conductivity and alkalinity that needs to be revived and reviewed, and in the process check how your contribution fits together. There are many definitions of alkalinity and equations proposed in the literature. We were trying to organize this into different models that the user could choose from (with parameter values, in the case of electrical conductivity, for example, provided via local or embedded param files). Also, the idea would be to produce alkalinity values using the same approach in PHREEQC (which would be the default model if not explicitly changed by the user), but it could be different (based on the definitions of Wolf-Gladrow et al.).
> 
> I suggest we keep this PR open for now.

Hi Allan,

Sounds great! Please lmk if you need help when you continue the past work, I can give you a hand.

Reaktoro/Utils/AqueousProps.cpp
```diff
@@ -388,8 +386,61 @@ struct AqueousProps::Impl
 
     auto alkalinity() const -> real
     {
-        error(true, "AqueousProps::alkalinity has not been implemented yet.");
-        return {};
+        // The number of species in the system
+        const Index num_species = system.species().size();
+
+        // The factors_and_species that contribute to alkalinity
+        static const auto factors_and_species = Pairs<double, std::string>{
```
allanleal: This can be moved to the constructor of `AqueousProps`. In the process, adapt this so that the indices of the species are saved along with their alkalinity factors -- instead of formulas, which will require finding indices in each alkalinity call. Something like this:

```c++
alkalinity_factors = {
    { 1.0, aqspecies.findWithFormula("Na+")   },
    { 2.0, aqspecies.findWithFormula("Mg+2")  },
    { 2.0, aqspecies.findWithFormula("Ca+2")  },
    { 1.0, aqspecies.findWithFormula("K+")    },
    { 2.0, aqspecies.findWithFormula("Sr+2")  },
    {-1.0, aqspecies.findWithFormula("Cl-")   },
    {-1.0, aqspecies.findWithFormula("Br-")   },
    {-1.0, aqspecies.findWithFormula("NO3-")  },
    {-1.0, aqspecies.findWithFormula("H3PO4") },
    {-1.0, aqspecies.findWithFormula("H2PO4-")},
    {-1.0, aqspecies.findWithFormula("HPO4-2")}, 
    {-1.0, aqspecies.findWithFormula("PO4-3") }, 
    { 1.0, aqspecies.findWithFormula("NH3")   },
    { 1.0, aqspecies.findWithFormula("NH4+")  },
    {-2.0, aqspecies.findWithFormula("SO4-2") },
    {-2.0, aqspecies.findWithFormula("HSO4-") }, 
    {-1.0, aqspecies.findWithFormula("F-")    },
    {-1.0, aqspecies.findWithFormula("HF")    },
    {-1.0, aqspecies.findWithFormula("NO2-")  },
    {-1.0, aqspecies.findWithFormula("HNO2")  },
};
```

Reaktoro/Utils/AqueousProps.cpp
```diff
@@ -388,8 +386,61 @@ struct AqueousProps::Impl
...
+            // Calculate contribution to alkalinity if the species is in the system
+            if (i < num_species) {
+                auto alkalinity_species_amount = molar_amounts[i];
+                alkalinity += alkalinity_factor * alkalinity_species_amount;
+            }
+        }
+
+        // Convert to eq/L
+        auto m3_to_liter = 1000.0;
+        auto V = props.volume();  // in m3
```
allanleal: `props.volume()` will return the volume of the chemical system. Not good if more than one phase, in addition to the aqueous solution, is considered in the system.

Reaktoro/Utils/AqueousProps.cpp
```diff
@@ -388,8 +386,61 @@ struct AqueousProps::Impl
...
+        };
+
+        // Iterate over all alkalinity factors and ions
+        auto molar_amounts = props.speciesAmounts();
+        real alkalinity = 0.0;
+        for(const auto& [alkalinity_factor, alkalinity_species_name] : factors_and_species)
+        {
+            // Get the index of the current ion
+            const auto& aq_species_list = system.species().withAggregateState(AggregateState::Aqueous);
+            const Index i = aq_species_list.findWithFormula(alkalinity_species_name);
```
allanleal: We should avoid evaluating the indices of the species in each call to `alkalinity`. See comment on initializing alkalinity factors and associated species indices and the constructor of `AqueousProps`.

Reaktoro/Utils/AqueousProps.py
```diff
@@ -23,3 +23,117 @@
 # TODO Implement tests for the python bindings of component AqueousProps in AqueousProps[test].py
 def testAqueousProps():
     pass
+
+
+@pytest.mark.parametrize(
+    "db, composition", (
+        (SupcrtDatabase("supcrt07"), {
```
allanleal: Here all compositions are the same; the only difference is the database. Best if we could have multiple compositions, covering different cases of alkalinity contribution.

Reaktoro/Utils/AqueousProps.py
```diff
@@ -23,3 +23,117 @@
...
+    state = ChemicalState(system)
+    state.temperature(25, "celsius")
+    state.pressure(1, "atm")
+    for species_name, amount_and_units in composition.items():
+        species_amount, species_unit = amount_and_units
+        state.set(species_name, species_amount, species_unit)
+
+    aq_props = AqueousProps(state)
+    # Within 0.1% of relative deviation compared with the paper
+    assert aq_props.pH().val() == pytest.approx(8.2, rel=1e-3)
```
allanleal: I'd propose tests in which we purely evaluate the calculated alkalinities against expected values (that we can calculate easily/alternatively). Will try to do this. 

Also, I don't see the need to check pH here. Best if confine these tests to alkalinity only; only considering its functional aspect.

Reaktoro/Utils/AqueousProps.py
```diff
@@ -23,3 +23,117 @@
...
+    # Within 2% of relative deviation compared with the paper
+    assert aq_props.alkalinity().val() == pytest.approx(2.33e-3, rel=2e-2)
+
+
+def testAlkalinity():
+    """
```
allanleal: This test here follows more or less the idea outlined in the previous comment. But we could have this done in a more general way so that multiple compositions can be tested.

allanleal (commented):
Hi Diego, I don't see the plans outline before (on a more flexible impl. of alkalinity models) taking place soon, so let's revive this implementation you kindly proposed here. I'll soon push a rebased version of this PR with changes that addresses the comments above.

allanleal:
I'll merge this, Diego, and make it available in the next release, v2.11.0, which I'm working on. Thanks a lot for contributing.

volpatto:
> I'll merge this, Diego, and make it available in the next release, v2.11.0, which I'm working on. Thanks a lot for contributing.

Thanks for the improvements, reviews, and support, @allanleal!

PR #348, opened 2023-12-13
Fix bug with static variables in two activity models

hendriko373:
This PR intents to fix a bug related to cache invalidation in the Duan-Sun and Rumpf activity models. In these models some variables where declared as static, but never updated. As a result, the cached values are not invalidated between sequential computations that make use of the activity models.

In a first attempt the variables were made non-static, resolving the bug. Since this gives up caching altogether, some efficiency might be lost-- for a specific use-case I noted a runtime that took two times longer. Therefore, caching for the variables is reintroduced, with the caching object created and stored next to the `charged` species list that it is supposed to be in sync with.

@allanleal Being a first-time contributor here, I might easily have missed better options. I am all ears for suggestions.

### DIFF
--- a/Reaktoro/Models/ActivityModels/ActivityModelDuanSun.cpp
+++ b/Reaktoro/Models/ActivityModels/ActivityModelDuanSun.cpp
@@ -88,6 +88,15 @@ auto ActivityModelDuanSun(String gas) -> ActivityModelGenerator
         // The index of the dissolved gas in the aqueous phase.
         const auto igas = species.indexWithFormula(gas);
 
+        // the local indices of some charged species among all charged species
+        const auto aqmix = AqueousMixture(species);
+        const auto iNa  = aqmix.charged().findWithFormula("Na+");
+        const auto iK   = aqmix.charged().findWithFormula("K+");
+        const auto iCa  = aqmix.charged().findWithFormula("Ca++");
+        const auto iMg  = aqmix.charged().findWithFormula("Mg++");
+        const auto iCl  = aqmix.charged().findWithFormula("Cl-");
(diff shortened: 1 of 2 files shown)


allanleal:
Hi @hendriko373 , I see the issue. The assumption was that these models would always be called after another that exported the `AqueousMixture` object. Perhaps a simpler fix would be moving the block of code which determines the species indices to the parent scope, above the lambda function, and then capture them by copy. This would require creating an auxiliary `AqueousMixture` object.

hendriko373:
> Hi @hendriko373 , I see the issue. The assumption was that these models would always be called after another that exported the `AqueousMixture` object. Perhaps a simpler fix would be moving the block of code which determines the species indices to the parent scope, above the lambda function, and then capture them by copy. This would require creating an auxiliary `AqueousMixture` object.

Hi Allan. Thanks for the suggestion. Is it possible that the auxiliary mixture and the one referenced to in the `ActivityPropsRef` object, passed to the `fn` lambda, get out of sync? If not, I am happy to adapt this PR with your suggestion.

hendriko373:
@allanleal I updated the branch, using the auxiliary mixture object, as you suggested.

hendriko373:
@allanleal In case there is anything I can help with to further this PR, please let me know.

allanleal:
~Hi Hendrik, thanks for reaching out. I'll try to revise this tomorrow morning, before I'm away until Thursday.~
It looks good. Once the ci checks below pass, I'll merge.

hendriko373:
> ~Hi Hendrik, thanks for reaching out. I'll try to revise this tomorrow morning, before I'm away until Thursday.~ It looks good. Once the ci checks below pass, I'll merge.

Great, thanks!

PR #354, opened 2024-01-08
Only install valgrind for linux in  environment.devenv.yml

cdeil:
For me on MacOS I had to comment out `valgrind` because no conda package exists.

Note that in `valgrind [unix]` the `unix` means `linux` or `macos`
https://conda-devenv.readthedocs.io/en/latest/usage.html

I think that valgrind is only supported on super-old MacOS ( see https://valgrind.org ) so effectively I don't think it can be used on MacOS. But I don't really know to be honest.


allanleal:
Thanks, Christoph!

PR #367, opened 2024-02-08
Improve SmartEquilibriumSolver

rafaeljalves:
I encountered some issues with `SmartKineticsSolver`, as demonstrated in [this](https://colab.research.google.com/github/rafaeljalves/example_notebooks/blob/main/smart_solvers_error.ipynb) example. The initial error occurred during a reactive transport simulation, and the conditions could not be entirely replicated in a 0d case. 
The changes below are intended to prevent this type of error.

- Avoid creating an equilibrium predictor if the learning step did not succeed.
- Avoid returning nan during Taylor approximation
- Create a backup state that is used if the prediction step fails(this is helpful in case a prediction fails due to negative amounts)

### DIFF
--- a/Reaktoro/Equilibrium/SmartEquilibriumSolver.cpp
+++ b/Reaktoro/Equilibrium/SmartEquilibriumSolver.cpp
@@ -92,6 +92,9 @@ struct SmartEquilibriumSolver::Impl
     auto solve(ChemicalState& state, EquilibriumConditions const& conditions) -> SmartEquilibriumResult
     {
         tic(SOLVE_STEP)
+                 
+        // Save a backup state in case the smart prediction fails.
+        const auto statebkp = state;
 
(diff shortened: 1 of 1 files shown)


allanleal:
Thanks, @rafaeljalves - it makes sense to use the original chemical state (before the prediction) as the initial estimate of the full calculation in case the former produces a spurious predicted state.

In some cases, the predicted state could, however, be a better initial estimate for the full calculation. I'll merge this PR and research on a way that permits the predicted state to be (safely) used as an initial estimate rather than the original backup state.

rafaeljalves:
> In some cases, the predicted state could, however, be a better initial estimate for the full calculation. I'll merge this PR and research on a way that permits the predicted state to be (safely) used as an initial estimate rather than the original backup state.

Great! That sounds really good!

PR #392, opened 2024-08-05
Fix invalid `phaseProps` results

tadeu:
Fix an object lifetime problem, the returned `ChemicalPropsPhaseRef` were being dropped by `pybind11`, which were assuming it was a reference due to `py::return_value_policy::reference_internal`. The result was that, after some other deallocations and allocations, the object methods were returning values from unowned memory.

`py::keep_alive<0, 1>()` lets `pybind11` manage the returned object and tie its lifetime to the owner.


allanleal:
Thanks for the fix @tadeu!

